PR #1, opened 2016-03-11
Added authentication classes

clintonb:
ECOM-3895

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,16 +8,12 @@ clean:
 	rm -rf build
 
 quality:
-	pep8 --config=.pep8 $(PACKAGE)
-	pylint --rcfile=pylintrc $(PACKAGE)
+	tox -e quality
 
 requirements:
 	pip install -r test_requirements.txt
 
 test:
-	django-admin.py test $(PACKAGE) --settings=test_settings --with-coverage --cover-package=$(PACKAGE)
-	coverage report
+	tox
 
-validate: clean test quality
-
-.PHONY: clean, quality, requirements, validate
+.PHONY: clean, quality, requirements
--- a/Makefile
+++ b/Makefile
@@ -15,10 +15,8 @@ requirements:
 	pip install -r test_requirements.txt
 
 test:
-	nosetests --with-coverage --cover-inclusive --cover-branches \
-		--cover-html --cover-html-dir=$(COVERAGE)/html/ \
-		--cover-xml --cover-xml-file=$(COVERAGE)/coverage.xml \
-		--cover-package=$(PACKAGE) $(PACKAGE)/
+	django-admin.py test $(PACKAGE) --settings=test_settings --with-coverage --cover-package=$(PACKAGE)
+	coverage report
 
 validate: clean test quality
 
--- a/edx_rest_framework_extensions/authentication.py
+++ b/edx_rest_framework_extensions/authentication.py
@@ -1,0 +1,84 @@
+""" Authentication classes. """
+
+import logging
+
+import requests
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from rest_framework import exceptions
+from rest_framework.authentication import get_authorization_header, BaseAuthentication
+from rest_framework.status import HTTP_200_OK
+
+logger = logging.getLogger(__name__)
+User = get_user_model()
+
+
+class BearerAuthentication(BaseAuthentication):
+    """
+    Simple token based authentication.
+
+    This authentication class is useful for authenticating an OAuth2 access token
+     against a remote authentication provider.
+
+    Clients should authenticate by passing the token key in the "Authorization"
+    HTTP header, prepended with the string "Bearer ".
+
+    Examples:
+        Authorization: Bearer 401f7ac837da42b97f613d789819ff93537bee6a
+    """
+
+    def get_access_token_url(self):
+        """ Returns the URL, hosted by the OAuth2 provider, against which access tokens can be validated. """
+        return getattr(settings, 'OAUTH2_ACCESS_TOKEN_URL', None)
+
+    def authenticate(self, request):
+        if not self.get_access_token_url():
+            return None
+
+        auth = get_authorization_header(request).split()
+
+        if not auth or auth[0].lower() != b'bearer':
+            return None
+
+        if len(auth) == 1:
+            raise exceptions.AuthenticationFailed('Invalid token header. No credentials provided.')
+        elif len(auth) > 2:
+            raise exceptions.AuthenticationFailed('Invalid token header. Token string should not contain spaces.')
+
+        return self.authenticate_credentials(auth[1].decode('utf8'))
+
+    def authenticate_credentials(self, token):
+        """
+        Validate the bearer token against the OAuth provider.
+
+        Arguments:
+            token (str) -- Access token to validate
+
+        Returns:
+            User -- User associated with the access token
+            str -- Access token
+
+        Raises:
+            AuthenticationFailed -- when user is inactive, or does not exist.
+        """
+
+        url = '{root}/{token}'.format(root=self.get_access_token_url().rstrip('/'), token=token)
+        response = requests.get(url)
+
+        if response.status_code != HTTP_200_OK:
+            raise exceptions.AuthenticationFailed('Invalid token.')
+
+        data = response.json()
+
+        try:
+            user = User.objects.get(username=data['username'])
+        except User.DoesNotExist:
+            raise exceptions.AuthenticationFailed('The user linked to this token does not exist on this system.')
+
+        if not user.is_active:
+            raise exceptions.AuthenticationFailed('User inactive or deleted.')
+
+        return user, token
+
+    def authenticate_header(self, request):
+        return 'Bearer'
--- a/edx_rest_framework_extensions/tests/test_authentication.py
+++ b/edx_rest_framework_extensions/tests/test_authentication.py
@@ -1,0 +1,128 @@
+# -*- coding: utf-8 -*-
+""" Tests for authentication classes. """
+
+import json
+
+import httpretty
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from django.test import override_settings, RequestFactory, TestCase
+from rest_framework.exceptions import AuthenticationFailed
+
+from edx_rest_framework_extensions.authentication import BearerAuthentication
+from edx_rest_framework_extensions.tests import factories
+
+OAUTH2_ACCESS_TOKEN_URL = 'http://example.com/oauth2/access_token/'
+
+User = get_user_model()
+
+
+class AccessTokenMixin(object):
+    """ Test mixin for dealing with OAuth2 access tokens. """
+    DEFAULT_TOKEN = 'abc123'
+
+    def mock_access_token_response(self, status=200, token=DEFAULT_TOKEN, username='fake-user'):
+        """ Mock the access token endpoint response of the OAuth2 provider. """
+        url = '{root}/{token}'.format(root=OAUTH2_ACCESS_TOKEN_URL.rstrip('/'), token=token)
+        httpretty.register_uri(
+            httpretty.GET,
+            url,
+            body=json.dumps({'username': username, 'scope': 'read', 'expires_in': 60}),
+            content_type="application/json",
+            status=status
+        )
+
+
+@override_settings(OAUTH2_ACCESS_TOKEN_URL=OAUTH2_ACCESS_TOKEN_URL)
+class BearerAuthenticationTests(AccessTokenMixin, TestCase):
+    """ Tests for the BearerAuthentication class. """
+
+    def setUp(self):
+        super(BearerAuthenticationTests, self).setUp()
+        self.auth = BearerAuthentication()
+        self.factory = RequestFactory()
+
+    def create_authenticated_request(self, token=AccessTokenMixin.DEFAULT_TOKEN, token_name='Bearer'):
+        """ Returns a Request with the authorization set using the specified values. """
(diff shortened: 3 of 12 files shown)


clintonb:
@bderusha please review the authentication class.
@nedbat please review the tox setup, or recommend someone else who can

FYI @edx/devops @nasthagiri @jcdyer @mattdrayer...we should get a tech leads/architects group.

clintonb:
FYI @edx/ecommerce. Sorry DevOps.

edx_rest_framework_extensions/authentication.py
```diff
@@ -0,0 +1,81 @@
...
+        return getattr(settings, 'OAUTH2_PROVIDER_URL', None)
+
+    def authenticate(self, request):
+        provider_url = self.get_provider_url()
+        if not provider_url:
+            return None
+
+        provider_url = provider_url.strip('/')
+
+        auth = get_authorization_header(request).split()
```
bderusha: As someone less familiar with the expected structure of this header string can we include a comment about what we are expecting to get back here.

It's not clear what an auth list contains, which is exacerbated further as this method continues to just index into the list to get what it needs.

edx_rest_framework_extensions/authentication.py
```diff
@@ -0,0 +1,81 @@
...
+
+        Returns:
+            User -- User associated with the access token
+            str -- Access token
+
+        Raises:
+            AuthenticationFailed -- when user is inactive, or does not exist.
+        """
+        try:
+            url = '{}/access_token/{}/'.format(provider_url, token)
```
bderusha: please use named variables here

edx_rest_framework_extensions/authentication.py
```diff
@@ -0,0 +1,81 @@
...
+            response = requests.get(url)
+            if response.status_code != HTTP_200_OK:
+                raise exceptions.AuthenticationFailed('Invalid token.')
+
+            data = response.json()
+            user = User.objects.get(username=data['username'])
```
bderusha: NIT: the `try`/`except` block really only applies to this line. Given that, can we just wrap this one call with it?

edx_rest_framework_extensions/authentication.py
clintonb: Se the docstring.

edx_rest_framework_extensions/authentication.py
```diff
@@ -0,0 +1,81 @@
...
+        except User.DoesNotExist:
+            raise exceptions.AuthenticationFailed('Invalid token.')
+
+        if not user.is_active:
+            raise exceptions.AuthenticationFailed('User inactive or deleted.')
+
+        return user, token
+
+    def authenticate_header(self, request):
```
bderusha: `request` is unused by this method.  Let's use `__` if this parameter is needed at all.

edx_rest_framework_extensions/tests/test_authentication.py
```diff
@@ -0,0 +1,113 @@
...
+User = get_user_model()
+
+
+class AccessTokenMixin(object):
+    """ Test mixin for dealing with OAuth2 access tokens. """
+    DEFAULT_TOKEN = 'abc123'
+
+    def mock_access_token_response(self, status=200, token=DEFAULT_TOKEN, username='fake-user'):
+        """ Mock the access token endpoint response of the OAuth2 provider. """
+        httpretty.register_uri(httpretty.GET, '{}/access_token/{}/'.format(OAUTH2_PROVIDER_URL, token),
```
bderusha: named variables please

edx_rest_framework_extensions/authentication.py
bderusha: ::EDIT::
Missed that this is in the doc string.  Still might be nice to name some of the pieces

edx_rest_framework_extensions/tests/test_authentication.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def create_authenticated_request(self, token=AccessTokenMixin.DEFAULT_TOKEN, token_name='Bearer'):
+        """ Returns a Request with the authorization set using the specified values. """
+        auth_header = '{token_name} {token}'.format(token_name=token_name, token=token)
+        request = self.factory.get('/', HTTP_AUTHORIZATION=auth_header)
+        return request
+
+    def test_authenticate_header(self):
+        """ The method should return the string Bearer. """
+        self.assertEqual(self.auth.authenticate_header(self.create_authenticated_request()), 'Bearer')
```
bderusha: As written this method doesn't need an authenticated request passed to it.  It is confusing to have all these references to a request being passed in if we don't really need it.  I'd prefer to see this not send any arguments in the test.

edx_rest_framework_extensions/authentication.py
clintonb: Updated.

edx_rest_framework_extensions/authentication.py
clintonb: This is overriding a method from the base class. I prefer to keep the same signature.

edx_rest_framework_extensions/tests/test_authentication.py
clintonb: Updated.

bderusha:
A couple of little things to clean up.  :+1: after that

edx_rest_framework_extensions/tests/test_authentication.py
clintonb: I prefer to maintain the original signature.

edx_rest_framework_extensions/authentication.py
clintonb: Updated.

tox.ini
```diff
@@ -0,0 +1,25 @@
...
+    -rtest_requirements.txt
+
+commands =
+    django-admin.py test edx_rest_framework_extensions --settings=test_settings --with-coverage --cover-package=edx_rest_framework_extensions
+    coverage report
+
+[testenv:quality]
+deps =
+    django18: Django>=1.8,<1.9
+    django19: Django>=1.9,<1.10
```
cpennington: Could you set these django dependencies up in the base `[tox]` config so that they apply to all environments?

edx_rest_framework_extensions/authentication.py
```diff
@@ -0,0 +1,87 @@
...
+    def get_provider_url(self):
+        """ Returns the root URL of the OAuth2 provider. """
+        return getattr(settings, 'OAUTH2_PROVIDER_URL', None)
+
+    def authenticate(self, request):
+        provider_url = self.get_provider_url()
+        if not provider_url:
+            return None
+
+        provider_url = provider_url.strip('/')
```
jcdyer: Stripping off the leading slashes from the provider URL introduces possibly unexpected behavior.  A site-relative absolute URL like `/path/to/endpoint` will be changed to the relative URL `path/to/url`, which has a very different meaning.  A protocol-relative URL like `//facebook.com/path/to/endpoint/` will be changed to the very different url `facebook.com/path/to/endpoint`.  Unless this is guarded against elsewhere, and specifically desired behavior for reasons I'm not seeing, I'd suggest using `provider_url.rstrip()` to only trim trailing slashes.

edx_rest_framework_extensions/authentication.py
jcdyer: I support keeping the same signature.  Does it not need a pylint disable pragma for an unused parameter?

tox.ini
clintonb: Almost. The `deps` value in `[testenv]` is inherited by `[testenv:*]`. Updated.

edx_rest_framework_extensions/tests/test_authentication.py
```diff
@@ -0,0 +1,117 @@
...
+User = get_user_model()
+
+
+class AccessTokenMixin(object):
+    """ Test mixin for dealing with OAuth2 access tokens. """
+    DEFAULT_TOKEN = 'abc123'
+
+    def mock_access_token_response(self, status=200, token=DEFAULT_TOKEN, username='fake-user'):
+        """ Mock the access token endpoint response of the OAuth2 provider. """
+        url = '{root}/access_token/{token}/'.format(root=OAUTH2_PROVIDER_URL, token=token)
```
jcdyer: Here, the provider URL isn't strip()ed or rstrip()ed at all.  Why the discrepancy?

edx_rest_framework_extensions/authentication.py
clintonb: We are using Pylint 1.5.4, which may do a better job of inspecting parent classes. Pylint is not complaining about this signature; however, it does complain if I add another argument.

tox.ini
```diff
@@ -0,0 +1,21 @@
...
+    PYTHONPATH = {toxinidir}
+
+deps =
+    django18: Django>=1.8,<1.9
+    django19: Django>=1.9,<1.10
+    -rtest_requirements.txt
+
+commands =
+    django-admin.py test edx_rest_framework_extensions --settings=test_settings --with-coverage --cover-package=edx_rest_framework_extensions
+    coverage report
```
nedbat: You might prefer to have one combined coverage report, instead of four individual ones.

tox.ini
```diff
@@ -0,0 +1,21 @@
+[tox]
+envlist = py{27,35}-django{18,19}, quality
```
stvstnfrd: You're adding py3 support here, which is great, but it doesn't look like you're actually running the tests on Travis with py3.

We should presumably add this test case to `.travis.yml` to ensure it is always run and we don't inadvertently introduce mistakes.

https://tox.readthedocs.org/en/latest/config.html#confval-basepython=NAME-OR-PATH

Makefile
```diff
@@ -8,18 +8,15 @@ clean:
...
 
 test:
-	nosetests --with-coverage --cover-inclusive --cover-branches \
-		--cover-html --cover-html-dir=$(COVERAGE)/html/ \
-		--cover-xml --cover-xml-file=$(COVERAGE)/coverage.xml \
-		--cover-package=$(PACKAGE) $(PACKAGE)/
+	tox
 
-validate: clean test quality
+validate:
```
stvstnfrd: At this point, it looks like the `test` and `validate` targets are identical: they just call out to `tox`.

The `validate` target should either be updated to contain unique logic or it should probably be removed entirely.

edx_rest_framework_extensions/authentication.py
clintonb: `provider_url` should always be a complete URL. Nevertheless `rstrip()` is appropriate.

edx_rest_framework_extensions/tests/test_authentication.py
clintonb: Updated.

tox.ini
stvstnfrd: +1. FWIW: When I've done this, I added a separate `coverage` testenv.
From there, the `coverage` testenv would run by default locally, but then I would instead run the  `coveralls` target on Travis (where the coveralls target both calculates coverage and also submits it to coveralls.io .

tox.ini
clintonb: We are running all combinations: https://travis-ci.org/edx/edx-drf-extensions/builds/116194344.

stvstnfrd: +1

Makefile
clintonb: Removed.

tox.ini
clintonb: @nedbat @stvstnfrd please provide examples. I have tried a number of different configurations, and cannot get coverage to run without having first run the tests. We are very near the point of diminishing returns—my working more on this to save 0.5s of runtime is not worthwhile—so, I am leaning toward improving this later.

clintonb:
All feedback addressed. Awaiting final comments from @nedbat and @stvstnfrd before merging.

tox.ini
stvstnfrd: @clintonb Yeah, you can't print a coverage report without first running the test.
My approach is basically:
- _only_ run tests in the default testenvs (`py{27,35}-django{18,19}`)
    - do not generate coverage stats or print a report
- add a separate testenv called `coverage` that runs tests and generates coverage

Here's what I have in another project.
I tried to remove unrelated work and add comments inline.

Hopefully that helps.

# tox.ini

```ini
# You can also add other tests here like `pylint`, `pep8`, etc.
[tox]
envlist = py27-dj18,coverage

# By default, just run tests normally.
# This will not generate coverage stats, nor will it generate a report.
[testenv]
commands = {envpython} manage.py test

# Typically, just run this locally.
# This should test code, generate coverage stats, and print a report.
[testenv:coverage]
deps =
    coverage
setenv =
    NOSE_COVER_TESTS=1
    NOSE_WITH_COVERAGE=1
commands =
    {envpython} manage.py test

# Only run this on Travis
# This should test code, generate coverage stats, and submit to coveralls.io
[testenv:coveralls]
deps =
    coverage
    coveralls
setenv =
    NOSE_COVER_TESTS=1
    NOSE_WITH_COVERAGE=1
passenv = TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH
commands =
    {envbindir}/coverage run --source=geoinfo manage.py test
    {envbindir}/coveralls
```

# .travis.yml

```yaml
script:
    - 'tox -v'
env:
    - TOXENV=py27-dj18
    - TOXENV=coveralls
```

# Makefile
```makefile
# The default `envlist` only contains `coverage`, not `coveralls`.
test:
    python manage.py test
```

clintonb: This doesn't work for codecov: https://travis-ci.org/edx/edx-drf-extensions/jobs/116245923. I'm going to move ahead with this as-is. We can improve it later.

PR #4, opened 2016-03-22
Validating OAuth2 token expiration date

clintonb:
BearerAuthentication has been updated to actually validate the expiration time of the access token response. This ensures that users with expired tokens are NOT authenticated.

ECOM-3895


clintonb:
@bderusha please review

bderusha:
:+1:

PR #13, opened 2016-06-22
Added support for JWT claim-user attribute mapping

clintonb:
This mapping will allow consuming projects to specify which claims should be mapped to specific user model attributes. By default the email claim is mapped to the email attribute, and the administrator claim is mapped to the is_staff attribute.

 ECOM-4793

### DIFF
--- a/edx_rest_framework_extensions/tests/test_authentication.py
+++ b/edx_rest_framework_extensions/tests/test_authentication.py
@@ -3,6 +3,7 @@
 import json
 from logging import Logger
 
+import ddt
 import httpretty
 import mock
 from django.contrib.auth import get_user_model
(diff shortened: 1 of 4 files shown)


clintonb:
@mjfrey @rlucioni please review

edx_rest_framework_extensions/tests/test_authentication.py
```diff
@@ -169,18 +170,28 @@ def test_authenticate_user_creation_with_request_exception(self):
...
     """ JWT Authentication class tests. """
-    def test_authenticate_credentials_user_creation(self):
-        """ Test whether the user model is being created and assigned fields from the payload. """
 
+    def get_jwt_payload(self, **additional_claims):
+        """ Returns a JWT payload with the necessary claims to create a new user. """
         email = 'gcostanza@gmail.com'
         username = 'gcostanza'
+        payload = additional_claims.copy()
+        payload.update({'preferred_username': username, 'email': email})
```
rlucioni: Can be done more concisely as:
```py
payload = dict({'preferred_username': username, 'email': email}, **additional_claims)
```

rlucioni:
👍

mjfrey:
👍 Thanks

edx_rest_framework_extensions/tests/test_authentication.py
clintonb: Nice. Updated.

PR #15, opened 2016-07-01
mattdrayer/catch-token-exception: Also handle InvalidTokenError

mattdrayer:
@clintonb @douglashall @e0d -- added InvalidTokenError to the list of exceptions we watch for during JWT decoding.

### DIFF
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '1.1.0'  # pragma: no cover
+__version__ = '1.1.1'  # pragma: no cover
--- a/edx_rest_framework_extensions/utils.py
+++ b/edx_rest_framework_extensions/utils.py
@@ -80,7 +80,7 @@ def jwt_decode_handler(token):
                 algorithms=[api_settings.JWT_ALGORITHM]
             )
             return decoded
-        except jwt.DecodeError:
+        except jwt.InvalidTokenError:
             msg = "Token decode failed for issuer '{issuer}'".format(issuer=jwt_issuer['ISSUER'])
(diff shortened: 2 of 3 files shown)


douglashall:
👍

edx_rest_framework_extensions/utils.py
```diff
@@ -80,7 +80,7 @@ def jwt_decode_handler(token):
                 algorithms=[api_settings.JWT_ALGORITHM]
             )
             return decoded
-        except jwt.DecodeError:
+        except (jwt.DecodeError, jwt.InvalidTokenError):
```
clintonb: Just `jwt.InvalidTokenError`. `jwt.DecodeError` is a subclass.

edx_rest_framework_extensions/__init__.py
```diff
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '1.1.0'  # pragma: no cover
+__version__ = '1.2.0'  # pragma: no cover
```
clintonb: 1.1.1

clintonb:
Tests?

edx_rest_framework_extensions/__init__.py
mattdrayer: Changed

edx_rest_framework_extensions/utils.py
mattdrayer: Changed

mattdrayer:
Tests are passing with coverage @ 100% without adding new tests?

mattdrayer:
I've added a new test which passes in an "invalid.token" to the decoding workflow

clintonb:
👍

mattdrayer:
Adding new release tag now

PR #18, opened 2016-12-21
Don't change warnings global settings

nedbat:
This line was causing test failures elsewhere in the edx-platform test suite, when warnings were appearing that shouldn't.
@clintonb @mattdrayer Do you know why this line is here? A library function like this shouldn't be changing the global warnings settings.
/cc @gsong

### DIFF
--- a/edx_rest_framework_extensions/settings.py
+++ b/edx_rest_framework_extensions/settings.py
@@ -56,7 +56,6 @@ def _get_deprecated_jwt_issuers():
     Having this allows for easier testing/mocking
     """
     # If JWT_ISSUERS is not defined, attempt to return the deprecated settings.
-    warnings.simplefilter('default')
     warnings.warn(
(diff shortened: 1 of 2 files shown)


nedbat:
@clintonb @mattdrayer Do you know why this line is here? A library function like this shouldn't be changing the global warnings settings.
/cc @gsong

mattdrayer:
It looks like I'm the one who added this line, but I can't recall any particular reason for doing so.  I believe it came about when adding test coverage for a DeprecationWarning workflow.

mattdrayer:
I guess I'm gonna do this...  👍

andy-armstrong:
I have to link to @nedbat's excellent blog post about how he diagnosed this problem. Amazing perseverance, Ned!

http://nedbatchelder.com//blog/201612/finding_test_coupling.html

PR #19, opened 2016-12-22
Bump version to 1.2.2

nedbat:


nedbat:
@clintonb This is all I have to do, then tag after merge to get it deployed, right?

PR #21, opened 2017-05-26
Added support for django 1.11

ahsan-ul-haq:
LEARNER-1182
Added support for Django 1.11 and Python 3.6.


edx_rest_framework_extensions/__init__.py
```diff
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '1.2.2'  # pragma: no cover
+__version__ = '1.3.0'  # pragma: no cover
```
clintonb: This is not worthy of a minor version bump. 1.2.3

tox.ini
```diff
@@ -10,6 +10,7 @@ deps =
     django18: Django>=1.8,<1.9
     django19: Django>=1.9,<1.10
     django110: Django>=1.10,<1.11
```
clintonb: This is probably why the tests are failing.

tox.ini
```diff
@@ -10,7 +10,7 @@ deps =
...
-    django110: Django>=1.11,<1.12
+    django111: Django>=1.11,<1.12
```
clintonb: There won't be a Django 1.12. Next version is 2.0.

ahsan-ul-haq:
@clintonb made the suggested changes

ahsan-ul-haq:
@clintonb made the change

PR #22, opened 2018-02-16
 Moved paginators from edx-platform for easy reuse

georgebabey:
Pagination should be standard and easy across all of edx applications.
By moving pagination into this package, it can be installed easily in
other apps/projects

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,9 +32,13 @@ setup(
     packages=['edx_rest_framework_extensions'],
     install_requires=[
         'django>=1.8.9,<2.0',
-        'djangorestframework>=3.2.3,<4.0.0',
+        'djangorestframework==3.6.3',
         'djangorestframework-jwt>=1.7.2,<2.0.0',
         'python-dateutil>=2.0',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -32,9 +32,13 @@
     packages=['edx_rest_framework_extensions'],
     install_requires=[
         'django>=1.8.9,<2.0',
-        'djangorestframework>=3.2.3,<4.0.0',
+        'djangorestframework==3.6.3',
```
georgebabey: It was necessary to pull in the edx fork of DRF as the latest version of DRF is not compatible with 1.9  This keeps this library in line with edx-platform

setup.py
```diff
@@ -32,9 +32,13 @@
...
         'djangorestframework-jwt>=1.7.2,<2.0.0',
         'python-dateutil>=2.0',
         'requests>=2.7.0,<3.0.0',
+        'six==1.11.0',
```
georgebabey: Needed for python3 compatibility (range in test_paginators.py)

PR #23, opened 2018-02-16
Moved paginators from edx-platform for easy reuse

georgebabey:
Pagination should be standard and easy across all of edx applications. By moving pagination into this package, it can be installed easily in other apps/projects

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,7 +3,7 @@ coverage:
   status:
     patch:
       default:
-        target: 95
+        target: 90
     project:
       default:
-        target: 95
+        target: 90
--- a/setup.py
+++ b/setup.py
@@ -32,9 +32,13 @@ setup(
     packages=['edx_rest_framework_extensions'],
     install_requires=[
         'django>=1.8.9,<2.0',
-        'djangorestframework>=3.2.3,<4.0.0',
+        'djangorestframework==3.6.3',
         'djangorestframework-jwt>=1.7.2,<2.0.0',
         'python-dateutil>=2.0',
         'requests>=2.7.0,<3.0.0',
+        'six==1.11.0',
     ],
+    dependency_links=[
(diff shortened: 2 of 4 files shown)


codecov.yml
```diff
@@ -3,7 +3,7 @@ coverage:
   status:
     patch:
       default:
-        target: 95
+        target: 90
```
georgebabey: edx-platform coverage target is 85% so this felt like a good compromise. This patch comes in at 94.91% with the tests moved from platform

setup.py
```diff
@@ -32,9 +32,13 @@
...
         'django>=1.8.9,<2.0',
-        'djangorestframework>=3.2.3,<4.0.0',
+        'djangorestframework==3.6.3',
         'djangorestframework-jwt>=1.7.2,<2.0.0',
         'python-dateutil>=2.0',
         'requests>=2.7.0,<3.0.0',
+        'six==1.11.0',
     ],
+    dependency_links=[
+        'git+https://github.com/edx/django-rest-framework.git@1ceda7c086fddffd1c440cc86856441bbf0bd9cb#egg=djangorestframework==3.6.3'
```
georgebabey: This pins us to the edx fork of DRF just as edx-platform has. Prior it was pulling in latest DRF which no longer supports 1.9

setup.py
```diff
@@ -32,9 +32,13 @@
     packages=['edx_rest_framework_extensions'],
     install_requires=[
         'django>=1.8.9,<2.0',
-        'djangorestframework>=3.2.3,<4.0.0',
+        'djangorestframework==3.6.3',
         'djangorestframework-jwt>=1.7.2,<2.0.0',
         'python-dateutil>=2.0',
         'requests>=2.7.0,<3.0.0',
+        'six==1.11.0',
```
georgebabey: Needed for python 3 compatibility in w/ range() in test_paginators.py

PR #25, opened 2018-03-07
WIP - Add ratelimiting django app.

adeelkhan:
This app would enable us to ratelimit
non drf based views. This patch provides mixin and
decorators (class method, function based) to enable
ratelimiting for a function, class or class method.

LEARNER-3858

### DIFF
--- a/edx_rest_framework_extensions/edx_ratelimit_app/admin.py
+++ b/edx_rest_framework_extensions/edx_ratelimit_app/admin.py
@@ -1,0 +1,11 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.contrib import admin
+from .models import EdxRateLimitConfiguration
+
+
+class EdxRateLimitAdmin(admin.ModelAdmin):
+    fields = ('request_frequency', 'time_window_duration', 'block', 'methods', 'ip_whitelist')
+
+admin.site.register(EdxRateLimitConfiguration, EdxRateLimitAdmin)
--- a/edx_rest_framework_extensions/edx_ratelimit_app/apps.py
+++ b/edx_rest_framework_extensions/edx_ratelimit_app/apps.py
@@ -1,0 +1,8 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.apps import AppConfig
+
+
+class EdxRatelimitAppConfig(AppConfig):
+    name = 'edx_ratelimit_app'
--- a/edx_rest_framework_extensions/edx_ratelimit_app/decorator.py
+++ b/edx_rest_framework_extensions/edx_ratelimit_app/decorator.py
@@ -1,0 +1,50 @@
+import logging
+from functools import wraps
+from django.http import HttpRequest, JsonResponse
+from ratelimit import ALL
+from ratelimit.utils import is_ratelimited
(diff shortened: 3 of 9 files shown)


nasthagiri:
FYI @feanil @ormsbee

feanil:
@adeelkhan I don't think the bug warrantes this change.  It's unclear from the PR what he shortcomnigs of the underlying django-ratelimit app are that we are trying to solve here.  But even if we do need to build our own things for this, I think storing the limits in the DB and looking them up each time is not scalable.  Can you provide more infor on the problem we're trying to solve that django-ratelimit can't handle?

adeelkhan:
@feanil If you go through the associated ticket [LEARNER-3858](https://openedx.atlassian.net/browse/LEARNER-3858), you would find answers to all these questions and why its needed. Summary is django-ratelimit package is not fulfilling some of the things that is needed for edx use case. New requirements are shared by @nasthagiri  and are posted in her [comments](https://openedx.atlassian.net/browse/LEARNER-3858?focusedCommentId=303615&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-303615). Secondly about scalability thing, yes i have already thought about it and mentioned this in my comments on the ticket [here](https://openedx.atlassian.net/browse/LEARNER-3858?focusedCommentId=310402&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-310402). I am looking to cache configurations for future access.

feanil:
@adeelkhan sorry I wasn't clear,  I did review the ticket and the reasoning was still unclear.  Nimisha had many suggestions that seem like better moves that moving to our own rate limit package.  In particular moving these endpoints to DRF.  Why not do that instead of making a new rate-limiting package.  I agree that @nasthagiri 's requests are good long term goals to have but I don't agree that this is the right approach.  In particular having these settings in the database will make it difficult to ensure that the same settnigs are on in mulitple environments.  We currently run 4 production environments and so making sure that rate limiting is synced across them now is a new burden that we don't have a clear answer for.

adeelkhan:
@feanil So let me try to clear this a bit. Goal here is to create something for long term purpose. We want to create a ratelimit package, that would help us achieve ratelimiting for non DRF class/function views.  The idea is we already have django-rate this would just be a wrapper on django-ratelimit package, but would have the added requirements. 

As i already posted moving to DRF has to be experimented/tested. Even if its easily doable, this seems to be effort for this single view and this is not scalable solution at all. There might be other non DRF based class or function based views that might also need ratelimiting. It's good to have a mixin, that could be configured to apply ratelimit for any non DRF views easily. 

I want to know a few things that are as follows.
1) Moving /oauth2/access_token to DRF means, we really want to move away from existing installed django-ratelimit package?
2) We are already using ratelimit package (django-ratelimit) for the end point /oauth2/access_token. Its working fine with no issues, whats the purpose of these added requirements? My understanding is we want to move to a package incorporating these requirements. Its like ratelimiting + requirements
3) Shouldn't we move now to long term goal? Why not create some thing that would help us apply ratelimit for any future view?
4) What is the right approach to keep configurations for ratelimiting? 
5) By environment i guess you mean IDA. Each environment would have its own configuration in db. Why you think we need to sync ratelimit configuration betweeen them? 

cc @raeeschachar @rock345

PR #28, opened 2018-06-06
Add DRF permission classes for enforcing OAuth2 scopes and filters.

douglashall:
This adds two new DRF permission classes which can be applied to DRF API views for enforcing OAuth2 [scopes](https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/oauth_dispatch/docs/decisions/0006-enforce-scopes-in-LMS-APIs.rst) and [filters](https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/oauth_dispatch/docs/decisions/0007-include-organizations-in-tokens.rst).

**Related PRs**
https://github.com/edx/edx-platform/pull/18317


edx_rest_framework_extensions/permissions.py
```diff
@@ -7,3 +10,65 @@ class IsSuperuser(BasePermission):
...
+    """
+
+    def has_permission(self, request, view):
+        jwt = request.auth
+
+        if not jwt:
+            return False
+
+        # Only validate scopes for restricted applications.
+        if not jwt.get('restricted'):
```
nasthagiri: Question: do we need this check?

douglashall:
@nasthagiri I think this is ready for final review. Can you give it a look when you get a chance? I'm working on tests for https://github.com/edx/edx-platform/pull/18317 now.

edx_rest_framework_extensions/permissions.py
```diff
@@ -1,9 +1,71 @@
...
+        required_scopes = set(self.get_scopes(request, view))
+        if required_scopes.issubset(jwt_scopes):
+            return True
+
+        if required_scopes:
+            # pylint: disable=attribute-defined-outside-init
+            self.message = {
+                "detail": PermissionDenied.default_detail,
+                "required_scopes": required_scopes,
+            }
```
nasthagiri: Does it work with `message` being a `dict`?

edx_rest_framework_extensions/permissions.py
```diff
@@ -1,9 +1,71 @@
...
+
+class JwtHasScope(TokenHasScope):
+    """
+    The request is authenticated as a user and the token used has the right scope.
+    """
+
+    @skip_unless_jwt_scopes_enforced
+    @jwt_authentication_required
+    def has_permission(self, request, view):
+        jwt_scopes = jwt_decode_handler(request.auth).get('scopes', [])
```
nasthagiri: For separation of concerns, does it make sense for all JWT token parsing to live in one python module?  So move the accessing of `scopes` from the token near `jwt_decode_handler`?

edx_rest_framework_extensions/permissions.py
```diff
@@ -1,9 +1,71 @@
...
+        """
+        Ensure that the course_id kwarg provided to the view contains one
+        of the organizations specified in the content provider filters
+        in the JWT used to authenticate.
+        """
+        course_key = CourseKey.from_string(view.kwargs.get('course_id'))
+        jwt_filters = jwt_decode_handler(request.auth).get('filters', [])
+        for jwt_filter in jwt_filters:
+            (provider_type, org_filter) = jwt_filter.split(':')
+            if provider_type == 'content_org' and org_filter == course_key.org:
```
nasthagiri: Same question as above, should the parsing of `filters` in the JWT token live with the other token parsing code?

edx_rest_framework_extensions/decorators.py
```diff
@@ -0,0 +1,40 @@
+""" Decorator functions. """
+from functools import wraps
+
+import waffle
+from edx_rest_framework_extensions.authentication import JwtAuthentication
+
+SWITCH_ENFORCE_JWT_SCOPES = 'oauth2.enforce_jwt_scopes'
```
nasthagiri: Taking lessons from edx-platform, it's useful to put all waffle declarations in a separate `config.py` file.

edx_rest_framework_extensions/tests/test_permission.py
```diff
@@ -27,3 +36,103 @@ def test_has_permission_with_invalid_users(self, user):
...
+        Test that the permission check returns true when scopes are validated.
+        """
+        waffle_mock.return_value = True
+        request = RequestFactory().get('/')
+        request.successful_authenticator = JwtAuthentication()
+        request.auth = generate_jwt(self.user, scopes=self.scopes)
+        view = Mock(required_scopes=self.scopes)
+        self.assertTrue(JwtHasScope().has_permission(request, view))
+
+    def test_scope_enforcement_disabled(self):
```
nasthagiri: Can you `ddt` this test - to make sure it returns `False` when scope enforcement is enabled?

edx_rest_framework_extensions/tests/test_permission.py
```diff
@@ -27,3 +36,103 @@ def test_has_permission_with_invalid_users(self, user):
...
+        """
+        Test that the permission check is skipped if scope enforcement is disabled.
+        """
+        request = RequestFactory().get('/')
+        self.assertTrue(JwtHasScope().has_permission(request, None))
+
+    @patch('edx_rest_framework_extensions.decorators.waffle.switch_is_active')
+    def test_not_jwt_authentication(self, waffle_mock):
```
nasthagiri: Nit: This test may read better and ensure coverage if there was a `ddt` of various values for `successful_authenticator` with their expected returns from `has_permission`.

edx_rest_framework_extensions/tests/test_permission.py
```diff
@@ -27,3 +36,103 @@ def test_has_permission_with_invalid_users(self, user):
...
+        self.scopes = ('grades:read', 'certificates:read')
+
+    @patch('edx_rest_framework_extensions.decorators.waffle.switch_is_active')
+    def test_scopes_validated(self, waffle_mock):
+        """
+        Test that the permission check returns true when scopes are validated.
+        """
+        waffle_mock.return_value = True
+        request = RequestFactory().get('/')
+        request.successful_authenticator = JwtAuthentication()
```
nasthagiri: Is there an end-to-end test to verify that the `request` object will always have `successful_authenticator` field set as this code expects?

edx_rest_framework_extensions/tests/test_permission.py
```diff
@@ -27,3 +36,103 @@ def test_has_permission_with_invalid_users(self, user):
...
+        request.auth = generate_jwt(self.user, scopes=[])
+        view = Mock(required_scopes=self.scopes)
+        self.assertFalse(JwtHasScope().has_permission(request, view))
+
+
+class JwtHasContentOrgFilterForRequestedCourseTests(TestCase):
```
nasthagiri: Nit: similar comment as above - regarding using `ddt` if possible for easier readability of tests and their coverage.

edx_rest_framework_extensions/permissions.py
```diff
@@ -1,9 +1,71 @@
...
+
 
 class IsSuperuser(BasePermission):
     """ Allows access only to superusers. """
 
     def has_permission(self, request, view):
         return request.user and request.user.is_superuser
+
+
+class JwtHasScope(TokenHasScope):
```
nasthagiri: Where would you be checking the version number on the JWT Token?  For forward compatibility, we want to:

1. Declare somewhere (in a class variable perhaps) that JWT_TOKEN_READ_VERSION = 1
2. Reject tokens whose major version number > JWT_TOKEN_READ_VERSION

This allows us to modify JWT token formats or semantics in the future while knowing that clients will reject tokens that they can't understand.

edx_rest_framework_extensions/permissions.py
```diff
@@ -1,9 +1,71 @@
 """ Permission classes. """
+from oauth2_provider.ext.rest_framework.permissions import TokenHasScope
```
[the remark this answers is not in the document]
douglashall: I think that is okay and should be expected given that this is just a customization of what the DOT permission class does.

edx_rest_framework_extensions/permissions.py
douglashall: I'm not understanding how that check relates to permissions or scopes. Wouldn't something like that be more appropriate in `JwtAuthentication`?

edx_rest_framework_extensions/permissions.py
douglashall: Meaning, add a new function to `edx_rest_framework_extensions.utils` that accepts the encoded JWT and returns the list of scopes?

edx_rest_framework_extensions/permissions.py
```diff
@@ -1,9 +1,71 @@
...
+    @skip_unless_jwt_scopes_enforced
+    @jwt_authentication_required
+    def has_permission(self, request, view):
+        jwt_scopes = jwt_decode_handler(request.auth).get('scopes', [])
+        required_scopes = set(self.get_scopes(request, view))
+        if required_scopes.issubset(jwt_scopes):
+            return True
+
+        if required_scopes:
+            # pylint: disable=attribute-defined-outside-init
```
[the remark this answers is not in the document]
douglashall: Awesome, thanks! I'll check to see if I can specify a dict doing it that way.

edx_rest_framework_extensions/permissions.py
douglashall: Yes, it does when we do it this way. I'll make sure it still works when declaring as a class variable.

edx_rest_framework_extensions/tests/test_permission.py
douglashall: Not that I'm aware of. Maybe in the DRF codebase itself? Is that something you think we should add to this package?

edx_rest_framework_extensions/permissions.py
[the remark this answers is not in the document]
douglashall: So, specifying the message as a class variable with work with the dict. The problem doing it that way though is that I wouldn't be able to specify the required scopes in the message because those are defined on the view class. I'm going to leave it this way both here and below for now.

douglashall:
@nasthagiri I have updated this based on your feedback. There are a couple of outstanding questions for your review. Thanks!

edx_rest_framework_extensions/tests/test_permission.py
```diff
@@ -27,3 +36,65 @@ def test_has_permission_with_invalid_users(self, user):
         request.user = user
         permission = IsSuperuser()
         self.assertFalse(permission.has_permission(request, None))
+
+
+@ddt.ddt
+class JwtHasScopeTests(TestCase):
```
nasthagiri: Can you add a test somewhere to make sure the `dict` object in the `message` field gets properly serialized and sent in the error response?

edx_rest_framework_extensions/permissions.py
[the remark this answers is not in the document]
nasthagiri: Right.  Although I was hoping that all Authentication Provider implementation can remain in a single Authentication service - rather than needing to be updated/maintained in all IDAs.  Anyway, I don't see an alternative unless we don't inherit from that class - so it's a tradeoff, which is fine for now.

edx_rest_framework_extensions/permissions.py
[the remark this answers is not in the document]
nasthagiri: I added a comment below about adding a test case for this.  I don't see any code in StackOverflow or in the `Permission` classes where they set request-specific `message` values.  Also, is there any downstream code that assumes `message` is a string?

So I'm unclear whether this way is supported.

edx_rest_framework_extensions/tests/test_permission.py
```diff
@@ -27,3 +36,65 @@ def test_has_permission_with_invalid_users(self, user):
...
+        (True, None, (), ('test:read'), True),
+        (False, JwtAuthentication(), (), ('test:read'), True),
+        (False, None, (), ('test:read'), True),
+    )
+    @ddt.unpack
+    @patch('edx_rest_framework_extensions.decorators.waffle.switch_is_active')
+    def test_has_permission(self, enforce_scopes, authentication_class, jwt_scopes,
+                            required_scopes, expected_result, waffle_mock):
+        """
+        Test that the permission check returns true when scopes are validated.
```
nasthagiri: true -> the expected result

edx_rest_framework_extensions/tests/test_permission.py
```diff
@@ -27,3 +36,65 @@ def test_has_permission_with_invalid_users(self, user):
...
+        (True, None, [], {'course_id': 'course-v1:edX+DemoX+Demo_Course'}, True),
+        (False, JwtAuthentication(), [], {'course_id': 'course-v1:edX+DemoX+Demo_Course'}, True),
+        (False, None, [], {'course_id': 'course-v1:edX+DemoX+Demo_Course'}, True),
+    )
+    @ddt.unpack
+    @patch('edx_rest_framework_extensions.decorators.waffle.switch_is_active')
+    def test_has_permission(self, enforce_scopes, authentication_class, jwt_filters,
+                            view_kwargs, expected_result, waffle_mock):
+        """
+        Test that the permission check returns true when scopes are validated.
```
nasthagiri: true -> the expected result

edx_rest_framework_extensions/permissions.py
```diff
@@ -1,9 +1,71 @@
...
+    def has_permission(self, request, view):
+        """
+        Ensure that the course_id kwarg provided to the view contains one
+        of the organizations specified in the content provider filters
+        in the JWT used to authenticate.
+        """
+        course_key = CourseKey.from_string(view.kwargs.get('course_id'))
+        jwt_filters = decode_jwt_filters(request.auth)
+        for jwt_filter in jwt_filters:
+            (provider_type, org_filter) = jwt_filter.split(':')
```
nasthagiri: Nit: I was thinking the parsing of filters would also live in the token parsing code module.

edx_rest_framework_extensions/permissions.py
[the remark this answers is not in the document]
douglashall: It is supported, see this DRF code: https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L56. I'll see if I can add a test case for the message.

edx_rest_framework_extensions/permissions.py
douglashall: Why do we need a version number for this? Wouldn't the parsing code just raise an exception if it isn't able to parse a particular value from the token?

edx_rest_framework_extensions/permissions.py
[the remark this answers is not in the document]
douglashall: Fair point. I'll just copy over the [get_scopes](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/contrib/rest_framework/permissions.py#L54) function from `TokenHasScope` and inherit from `BasePermission`.

edx_rest_framework_extensions/permissions.py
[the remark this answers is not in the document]
douglashall: On second thought, even though this will work it seems weird to be modifying a class variable from an instance method. I'm going to stick with a generic missing required scopes/filter message defined at the class level.

douglashall:
@nasthagiri I think this is ready for final review.

edx_rest_framework_extensions/utils.py
```diff
@@ -87,3 +91,32 @@ def jwt_decode_handler(token):
...
+    """
+    Verify that the JWT version is supported.
+    """
+    supported_version = semantic_version.Version(
+        settings.JWT_AUTH.get('JWT_SUPPORTED_VERSION', DEFAULT_JWT_SUPPORTED_VERSION)
+    )
+    jwt_version = semantic_version.Version(
+        decoded_token.get('version', supported_version)
+    )
+    if jwt_version > supported_version:
```
nasthagiri: Change this to compare the major version number only.  Thanks.

PR #29, opened 2018-06-11
adding openedx.yaml

mduboseedx:
This is only adding the openedx.yaml file that our bot uses to help manage OSPRs. Just for transparency, tests did not completely pass:

  py27-django18: commands succeeded
  py27-django19: commands succeeded
  py27-django110: commands succeeded
  py27-django111: commands succeeded
  py35-django18: commands succeeded
  py35-django19: commands succeeded
  py35-django110: commands succeeded
  py35-django111: commands succeeded
  py36-django111: commands succeeded
ERROR:   quality: commands failed
  docs: commands succeeded

Let me know if that is an issue.

### DIFF
--- a/openedx.yaml
+++ b/openedx.yaml
@@ -1,0 +1,8 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oep-0002-bp-repo-metadata.html#specification
+
+nick: drfx
+oeps: {}
+openedx-release: {ref: release}
+owner: edx/platform-team
+track-pulls: true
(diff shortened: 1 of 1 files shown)


openedx.yaml
```diff
@@ -0,0 +1,8 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oep-0002-bp-repo-metadata.html#specification
+
+nick: drfx
+oeps: {}
+openedx-release: {ref: release}
```
jmbowman: I don't think we want `openedx-release` set for repos representing packages that are installed in Open edX services.  It could be confusing if we tagged the latest master at the time of the release as `open-release/hawthorn`, but the version actually used in Hawthorn was an earlier commit released to PyPI.

PR #30, opened 2018-06-19
Middleware to ensure JwtHasScope permission for JWT Authentication

nasthagiri:
This PR introduces a Middleware as a [runway](http://www.scaledagileframework.com/architectural-runway/) for [enforcing OAuth scopes](https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/oauth_dispatch/docs/decisions/0006-enforce-scopes-in-LMS-APIs.rst).

The Middleware updates any REST API views that use `JwtAuthentication` to always use the new JwtHasScope permission class.

This is needed because the Django framework doesn't allow us to configure `Permission` classes that *must* be enforced by all endpoints.  Although it supports a [DEFAULT_PERMISSION_CLASSES](http://www.django-rest-framework.org/api-guide/permissions/#setting-the-permission-policy) setting, any endpoint can choose to override and not use the defaults.  By having a middleware, we can update permission classes for all endpoints at runtime.

In the future, we may also introduce a decorator for all endpoints to use with a static linter that provides more immediate feedback to developers - rather than updating permissions behind-the-scenes with a middleware.

### DIFF
--- a/edx_rest_framework_extensions/middleware.py
+++ b/edx_rest_framework_extensions/middleware.py
@@ -1,0 +1,46 @@
+"""
+Middleware to ensure best practices of DRF endpoints.
+"""
+import logging
+
+from rest_framework_jwt.authentication import BaseJSONWebTokenAuthentication
+
+from .permissions import JwtHasScope
+
+log = logging.getLogger(__name__)
+
+
+class EnsureJWTAuthSettingsMiddleware(object):
+    """
+    Django middleware object that ensures the proper Permission classes
+    are set on all endpoints that use JWTAuthentication.
+    """
+    _required_permission_classes = (JwtHasScope,)
+
+    def _includes_base_class(self, iter_classes, base_class):
+        """
+        Returns whether any class in iter_class is a subclass of the given base_class.
+        """
+        return any(
+            issubclass(auth_class, base_class) for auth_class in iter_classes,
+        )
+
+    def process_view(self, request, view_func, view_args, view_kwargs):  # pylint: disable=unused-argument
+        view_class = getattr(view_func, 'view_class', view_func)
+
+        view_authentication_classes = getattr(view_class, 'authentication_classes', tuple())
+        if self._includes_base_class(view_authentication_classes, BaseJSONWebTokenAuthentication):
+
+            view_permission_classes = getattr(view_class, 'permission_classes', tuple())
+
+            for perm_class in self._required_permission_classes:
+
+                if not self._includes_base_class(view_permission_classes, perm_class):
+                    log.warning(
+                        u"The view %s allows Jwt Authentication but needs to include the %s permission class.",
+                        view_class.__name__,
+                        perm_class.__name__,
+                    )
+
+            view_class.permission_classes = view_permission_classes
+            view_class.permission_classes += self._required_permission_classes
--- a/edx_rest_framework_extensions/tests/test_middleware.py
+++ b/edx_rest_framework_extensions/tests/test_middleware.py
@@ -1,0 +1,99 @@
+"""
+Unit tests for safe_endpoints Middleware.
+"""
+import ddt
+from itertools import product
+from mock import patch
+
+from django.test import TestCase, RequestFactory
+from rest_framework.authentication import SessionAuthentication
+from rest_framework_jwt.authentication import BaseJSONWebTokenAuthentication
+from rest_framework.decorators import api_view
+from rest_framework.views import APIView
+
+from ..middleware import EnsureJWTAuthSettingsMiddleware
+from ..permissions import JwtHasScope
+
+
+class SomeIncludedPermissionClass(object):
+    pass
+
+
+class SomeJwtAuthenticationSubclass(BaseJSONWebTokenAuthentication):
+    pass
+
+
+def some_auth_decorator(include_jwt_auth=True, include_required_perm=True):
+    def _decorator(f):
+        f.permission_classes = (SomeIncludedPermissionClass,)
+        f.authentication_classes = (SessionAuthentication,)
+        if include_jwt_auth:
+            f.authentication_classes += (SomeJwtAuthenticationSubclass,)
+        if include_required_perm:
+            f.permission_classes += (JwtHasScope,)
+        return f
+    return _decorator
+
+
+@ddt.ddt
+class TestEnsureJWTAuthSettingsMiddleware(TestCase):
+    def setUp(self):
+        super(TestEnsureJWTAuthSettingsMiddleware, self).setUp()
+        self.request = RequestFactory().get('/')
+        self.middleware = EnsureJWTAuthSettingsMiddleware()
+
+    def _assert_included(self, item, iterator, should_be_included):
+        if should_be_included:
+            self.assertIn(item, iterator)
+        else:
+            self.assertNotIn(item, iterator)
+
+    @ddt.data(
+        *product(
+            (True, False),
+            (True, False),
+            (True, False),
+        )
+    )
+    @ddt.unpack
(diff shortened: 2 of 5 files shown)


nasthagiri:
An earlier version of this PR was reviewed at https://github.com/edx/edx-platform/pull/18399.

edx_rest_framework_extensions/middleware.py
```diff
@@ -0,0 +1,46 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class EnsureJWTAuthSettingsMiddleware(object):
+    """
+    Django middleware object that ensures the proper Permission classes
+    are set on all endpoints that use JWTAuthentication.
+    """
+    _required_permission_classes = (JwtHasScope,)
```
nasthagiri: Note change from edx-platform PR: changed to the actual permission class.

edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -0,0 +1,83 @@
...
+
+from ..middleware import EnsureJWTAuthSettingsMiddleware
+from ..permissions import JwtHasScope
+
+
+class SomeIncludedPermissionClass(object):
+    pass
+
+
+class SomeJwtAuthenticationSubclass(BaseJSONWebTokenAuthentication):
```
nasthagiri: Note change from edx-platform PR: added an explicit subclass of `BaseJSONWebTokenAuthentication` to clarify the testing.

edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -0,0 +1,83 @@
...
+from rest_framework.authentication import SessionAuthentication
+from rest_framework_jwt.authentication import BaseJSONWebTokenAuthentication
+from rest_framework.decorators import api_view
+from rest_framework.views import APIView
+
+from ..middleware import EnsureJWTAuthSettingsMiddleware
+from ..permissions import JwtHasScope
+
+
+class SomeIncludedPermissionClass(object):
```
nasthagiri: Note change from edx-platform PR: renamed to be consistent with the Auth class below.

edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -0,0 +1,83 @@
...
+    pass
+
+
+class SomeJwtAuthenticationSubclass(BaseJSONWebTokenAuthentication):
+    pass
+
+
+def some_auth_decorator(include_jwt_auth=True, include_required_perm=True):
```
nasthagiri: Note change from edx-platform PR: renamed to be consistent with the Auth class above.

edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -0,0 +1,83 @@
...
+            f.permission_classes += (JwtHasScope,)
+        return f
+    return _decorator
+
+
+@ddt.ddt
+class TestEnsureJWTAuthSettingsMiddleware(TestCase):
+    def setUp(self):
+        super(TestEnsureJWTAuthSettingsMiddleware, self).setUp()
+        self.request = RequestFactory().get('/')
```
nasthagiri: Note change from edx-platform PR: using `RequestFactory` instead of `get_mock_request ` that was within an openedx/core folder.

edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -0,0 +1,93 @@
...
+            should_be_included=include_required_perm or include_jwt_auth,
+        )
+
+    def test_simple_view(self):
+        def some_simple_view(request):
+            pass
+
+        self.assertIsNone(
+            self.middleware.process_view(self.request, some_simple_view, None, None)
+        )
```
robrap: Should we add an assertion that the simple view doesn't have `view_class.permission_classes`?

edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -0,0 +1,93 @@
...
+
+class SomeIncludedPermissionClass(object):
+    pass
+
+
+class SomeJwtAuthenticationSubclass(BaseJSONWebTokenAuthentication):
+    pass
+
+
+def some_auth_decorator(include_jwt_auth=True, include_required_perm=True):
```
robrap: Nit: If you are relying on values being explicitly declared, you can drop the default values.  Unless you are using that as a pseudo-comment about type?

edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -0,0 +1,93 @@
...
+@ddt.ddt
+class TestEnsureJWTAuthSettingsMiddleware(TestCase):
+    def setUp(self):
+        super(TestEnsureJWTAuthSettingsMiddleware, self).setUp()
+        self.request = RequestFactory().get('/')
+        self.middleware = EnsureJWTAuthSettingsMiddleware()
+
+    def _assert_included(self, item, iterator, should_be_included):
+        if should_be_included:
+            self.assertIn(item, iterator)
```
robrap: Maybe doesn't matter, but to we want to ensure it is only there once?  In other words, that we aren't adding again in a place where it is already there?

edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -0,0 +1,93 @@
...
+
+    @ddt.data(
+        *product(
+            (True, False),
+            (True, False),
+            (True, False),
+        )
+    )
+    @ddt.unpack
+    def test_api_views(self, use_function_view, include_jwt_auth, include_required_perm):
```
robrap: You could also test that logging is happening when you expect it to.

edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -0,0 +1,93 @@
...
+from itertools import product
+
+from django.test import TestCase, RequestFactory
+from rest_framework.authentication import SessionAuthentication
+from rest_framework_jwt.authentication import BaseJSONWebTokenAuthentication
+from rest_framework.decorators import api_view
+from rest_framework.views import APIView
+
+from ..middleware import EnsureJWTAuthSettingsMiddleware
+from ..permissions import JwtHasScope
```
robrap: There could be an integration test somewhere for all of this.  But a manual test once this gets into edx-platform will probably suffice.

edx_rest_framework_extensions/middleware.py
```diff
@@ -0,0 +1,46 @@
...
+            for perm_class in self._required_permission_classes:
+
+                if not self._includes_base_class(view_permission_classes, perm_class):
+                    log.warning(
+                        u"The view %s allows Jwt Authentication but needs to include the %s permission class.",
+                        view_class.__name__,
+                        perm_class.__name__,
+                    )
+
+            view_class.permission_classes = view_class.permission_classes or tuple()
```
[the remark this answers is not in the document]
nasthagiri: yes. updated.

edx_rest_framework_extensions/tests/test_middleware.py
nasthagiri: I put in defaults since not all future callers need to pass in values for those optional parameters.

edx_rest_framework_extensions/tests/test_middleware.py
nasthagiri: I put this test in there to make sure we don't cause a 500 because of expectations on views that don't have an `api_view` decorator.  I'll add a comment to explain why this test exists.

edx_rest_framework_extensions/tests/test_middleware.py
nasthagiri: true.  updated.

PR #33, opened 2018-06-20
Fix bug related to use of DEFAULT_JWT_SUPPORTED_VERSION.

douglashall:


edx_rest_framework_extensions/__init__.py
```diff
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '1.3.0'  # pragma: no cover
+__version__ = '1.4.0'  # pragma: no cover
```
nasthagiri: Thanks for updating this.

PR #34, opened 2018-06-21
OAuth Scopes: additional JWT middleware and permissions support

nasthagiri:


edx_rest_framework_extensions/decorators.py
```diff
@@ -16,8 +16,9 @@ def skip_unless_jwt_authenticated(f):
     def decorated_function(*args, **kwargs):
         """ Determine if JwtAuthentication was used to authenticate the request. """
         request = args[1]
-        if isinstance(request.successful_authenticator, JwtAuthentication):
-            if getattr(request, 'auth', None):
+        is_jwt_auth, auth_successful = is_jwt_authenticated(request)
+        if is_jwt_auth:
```
robrap: If you are ok with my above comment, this becomes something like:
```
if is_successful_jwt_authentication(request):
   return f(*args, **kwargs)
elif is_bad_jwt_authentication(request):
   return False
...
```

edx_rest_framework_extensions/authentication.py
```diff
@@ -212,3 +212,9 @@ def authenticate_credentials(self, payload):
                 raise exceptions.AuthenticationFailed(msg)
 
         return user
+
+
+def is_jwt_authenticated(request):
```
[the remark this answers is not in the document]
nasthagiri: @douglashall You originally wrote the decorator for `skip_unless_jwt_authenticated`.  Is there a reason why `request.successful_authenticator` would be set but `request.auth` would not be?  Were you just being extra careful?  Or is this an expected condition?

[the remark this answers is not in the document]
douglashall: I was being extra cautious, it is not an expected condition.

edx_rest_framework_extensions/decorators.py
```diff
@@ -2,8 +2,8 @@
 from functools import wraps
 
 import waffle
-from edx_rest_framework_extensions.authentication import JwtAuthentication
 from edx_rest_framework_extensions.config import SWITCH_ENFORCE_JWT_SCOPES
+from edx_rest_framework_extensions.authentication import is_jwt_authenticated
 
 
 def skip_unless_jwt_authenticated(f):
```
douglashall: Should we get rid of this if the plan is to use `rest_condition`?

edx_rest_framework_extensions/authentication.py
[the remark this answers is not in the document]
nasthagiri: Ok.  I simplified the logic then.
Worst case, we would be enforcing the permission if there was an unexpected error - which is the safer route.

edx_rest_framework_extensions/jwt_decoder.py
```diff
@@ -0,0 +1,122 @@
+""" Utility functions. """
```
nasthagiri: **Note:** Not sure why git is treating this as new file.  This was renamed from utils.py to jwt_decoder.py.

edx_rest_framework_extensions/authentication.py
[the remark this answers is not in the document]
robrap: If you wanted, you could add an assertion that auth isn’t None. Or we could forget about it.

edx_rest_framework_extensions/decorators.py
douglashall: I was just thinking that if users of the permission classes decided to do something like:
```
permission_classes = [And(IsJwtAuthenticated, JwtHasScope)]
```
then we would be running the `is_jwt_authenticated` logic twice, but that probably isn't an issue.

edx_rest_framework_extensions/authentication.py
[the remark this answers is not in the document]
douglashall: Yes, let's add that assertion. Thanks!

edx_rest_framework_extensions/permissions.py
```diff
@@ -18,15 +16,44 @@ def has_permission(self, request, view):
...
+
+class IsUserInUrl(BasePermission):
+    """
+    Allows access if the the requesting user matches the user in the URL.
+    """
+    def has_permission(self, request, view):
+        user_parameter_name = 'username'
+        url_username = (
+            request.GET.get(user_parameter_name, '') or
+            getattr(request, 'parser_context', {}).get('kwargs', {}).get(user_parameter_name, '')
```
robrap: This might be silly, but if a url had both 'username' and this custom `user_parameter_name`, shouldn't this one take precedence?

edx_rest_framework_extensions/permissions.py
```diff
@@ -18,15 +16,44 @@ def has_permission(self, request, view):
...
+    """
+    Allows access to "global" staff users..
+    """
+    def has_permission(self, request, view):
+        return request.user.is_staff
+
+
+class IsUserInUrl(BasePermission):
+    """
+    Allows access if the the requesting user matches the user in the URL.
```
robrap: I'll fix the double-the.

edx_rest_framework_extensions/middleware.py
```diff
@@ -16,6 +16,7 @@ class EnsureJWTAuthSettingsMiddleware(object):
     are set on all endpoints that use JWTAuthentication.
     """
     _required_permission_classes = (JwtHasScope,)
+    _view_does_not_support_scope = 'VIEW_DOES_NOT_SUPPORT_SCOPES'
```
robrap: Making plural.

edx_rest_framework_extensions/permissions.py
```diff
@@ -18,15 +16,44 @@ def has_permission(self, request, view):
...
+
+class IsJwtAuthenticated(BasePermission):
+    """
+    Returns whether the request was successfully authenticated with JwtAuthentication.
+    """
+    def has_permission(self, request, view):
+        return is_jwt_authenticated(request)
+
+
 class JwtHasScope(BasePermission):
```
robrap: I think I'd like to rename this to `HasJwtScope`.  I think starting with `Has` is more consistent with the ones that start with `Is`.

edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -53,40 +58,67 @@ def _assert_included(self, item, iterator, should_be_included):
...
         view_class = view.view_class if use_function_view else view
 
+        # verify pre-conditions
         self._assert_included(
             SomeJwtAuthenticationSubclass,
             view_class.authentication_classes,
             should_be_included=include_jwt_auth,
         )
 
+        if not use_function_view:
```
robrap: Adding a comment like the one below.

edx_rest_framework_extensions/tests/test_permission.py
```diff
@@ -27,18 +25,60 @@ def test_has_permission(self, has_permission):
...
+@ddt.ddt
+class IsUserInUrlTests(TestCase):
+    """ Tests for the IsUserInUrl permission class. """
+
+    @ddt.data(True, False)
+    def test_has_permission(self, has_user_in_url):
+        user = factories.UserFactory(username='foo')
+        url = '/'
+        if has_user_in_url:
+            url += '?username={}'.format(user.username)
```
robrap: Do we need a test for a custom query param for username?

edx_rest_framework_extensions/permissions.py
douglashall: I think with the other permissions the subject is the permission class. With this one, the subject is the JWT, hence `JwtHasScope`.

edx_rest_framework_extensions/permissions.py
douglashall: Not sure I follow. What do you mean by "custom" `user_parameter_name`? I think using the variable is just a convenience so that the literal string does not have to be typed out twice in this function right?

edx_rest_framework_extensions/tests/test_permission.py
douglashall: See comment above.

edx_rest_framework_extensions/permissions.py
robrap: Leaving as-is.  Thanks.

edx_rest_framework_extensions/authentication.py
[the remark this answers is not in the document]
robrap: Done.

edx_rest_framework_extensions/permissions.py
robrap: Um. Yes. I don't know how my brain refactored that code.  Ignore this comment. :)

edx_rest_framework_extensions/tests/test_permission.py
robrap: Please ignore. :)

robrap:
Thanks @douglashall.  Here are all of my changes.  Let me know if all is well and I'll squash and merge.

PR #35, opened 2018-06-25
Handle ConditionalPermission in middleware.

ormsbee:
EnsureJWTAuthSettingsMiddleware's check for permissions classes would
fail if it was applied to a view that uses conditional permissions from
the rest_condition library, like:

    C(permissions.IsStaff) | permissions.IsUserInUrl

The cause of failure is that we were checking for subclasses of
particular permissions like IsStaff, and ConditionalPermissions are
outside of that hierarchy (and aren't even classes). This is one of the
contributing issues in ARCH-153.

This commit checks to see if a permission is a ConditionalPermission
and then expands it out so that we capture only the Permission classes
inside. So the check will pass as long as you have the required
permission class _somewhere_ in your view's permissions, with the
assumption being that if you're using the required permission in a
Conditional, you probably know what you're doing and we shouldn't
blindly append the required permission class at the end.

### DIFF
--- a/edx_rest_framework_extensions/middleware.py
+++ b/edx_rest_framework_extensions/middleware.py
@@ -31,19 +31,33 @@ class EnsureJWTAuthSettingsMiddleware(object):
         Adds permissions classes that should exist for Jwt based authentication,
         if needed.
         """
-        view_permission_classes = getattr(view_class, 'permission_classes', tuple())
+        view_permissions = list(getattr(view_class, 'permission_classes', []))
+
+        # Not all permissions are classes, some will be ConditionalPermission
+        # objects from the rest_condition library. So we have to crawl all those
+        # and expand them to see if our target classes are inside the
+        # conditionals somewhere.
+        permission_classes = []
+        classes_to_add = []
+        while view_permissions:
+            permission = view_permissions.pop()
+            if not hasattr(permission, 'perms_or_conds'):
+                permission_classes.append(permission)
+            else:
+                for child in getattr(permission, 'perms_or_conds', []):
+                    view_permissions.append(child)
 
         for perm_class in self._required_permission_classes:
-
-            if not self._includes_base_class(view_permission_classes, perm_class):
+            if not self._includes_base_class(permission_classes, perm_class):
                 log.warning(
-                    u"The view %s allows Jwt Authentication but needs to include the %s permission class.",
+                    u"The view %s allows Jwt Authentication but needs to include the %s permission class (adding it for you)",
                     view_class.__name__,
                     perm_class.__name__,
                 )
+                classes_to_add.append(perm_class)
 
-        view_class.permission_classes = view_permission_classes
-        view_class.permission_classes += self._required_permission_classes
+        if classes_to_add:
+            view_class.permission_classes += tuple(classes_to_add)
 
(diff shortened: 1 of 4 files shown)


edx_rest_framework_extensions/middleware.py
```diff
@@ -31,19 +31,29 @@ def _add_missing_jwt_permission_classes(self, view_class):
...
                 log.warning(
-                    u"The view %s allows Jwt Authentication but needs to include the %s permission class.",
+                    u"The view %s allows Jwt Authentication but needs to include the %s permission class (adding it for you)",
                     view_class.__name__,
                     perm_class.__name__,
                 )
-
-        view_class.permission_classes = view_permission_classes
-        view_class.permission_classes += self._required_permission_classes
+                view_class.permission_classes += self._required_permission_classes
```
ormsbee: Self note: we should append the missing classes one at a time, not blindly append the list.

edx_rest_framework_extensions/middleware.py
```diff
@@ -31,19 +31,29 @@ def _add_missing_jwt_permission_classes(self, view_class):
...
+        # objects from the rest_condition library. So we have to crawl all those
+        # and expand them to see if our target classes are inside the
+        # conditionals somewhere.
+        permission_classes = []
+        while view_permissions:
+            permission = view_permissions.pop()
+            if not hasattr(permission, 'perms_or_conds'):
+                permission_classes.append(permission)
+            else:
+                for child in getattr(permission, 'perms_or_conds', []):
```
nasthagiri: Is there any public API or interface on `ConditionalPermission` that we can use?  Do you think this assumes an implementation of it that may change over time?  Or is this part of its contract?

edx_rest_framework_extensions/middleware.py
nasthagiri: Is this so we prevent duplicates?  And also in case the permission existed in ConditionalPermission?

ormsbee: Yes.

edx_rest_framework_extensions/middleware.py
ormsbee: It's a public attribute. I don't know if it's really promised in their API, as their docs are pretty minimal (just a usage in their repo's README. But I'm adding the test now, so if they break compatibility in a later version, builds for this repo should start failing.

edx_rest_framework_extensions/middleware.py
ormsbee: Expanded version for others who might be watching: moving it into the inner loop like this means that we're no longer always appending it at the end. The assumption being that if you mention `JwtScope` somewhere, even nested in a conditional, then you've taken it into account and you know what you're doing, so we're not going to append it at the end since it will almost surely ruin any logic you meant to put in there.

But when I first moved it, I left the code the same, which means I would be appending the whole list of required permissions every time a missing permission was found. Which is boolean logic-wise a no-op (ANDing the same condition twice), and we currently only have one required permission -- but it's still redundant and bad practice.

ormsbee:
@nasthagiri, @douglashall: Ready for review

ormsbee:
Squashed commits. Will merge as soon as the build turns green.

PR #36, opened 2018-06-27
Update Permissions and JWT handling for OAuth Scopes

nasthagiri:
In addition to minor improvements, this PR does the following:

- Adds a design pattern for centrally managing token versions and updates in jwt_decoder.py via `JwtTokenVersion` and `_upgrade_token`.
- Updates the `EnsureJWTAuthSettingsMiddleware` so it no longer adds `required_scopes` to Views.
Instead, the `JwtHasScope` Permission will fail if the view doesn't have `required_scopes`.
- Adds a `JwtHasUserFilterForRequestedUser` Permission to verify the `user:me` filter.
- Adds a `JWT_RESTRICTED_APPLICATION_OR_USER_ACCESS` Permission constant that encapsulates checking for Jwt related permissions for APIs that have course-related and user-related data.

### DIFF
--- a/edx_rest_framework_extensions/config.py
+++ b/edx_rest_framework_extensions/config.py
@@ -2,4 +2,6 @@
 Application configuration constants and code.
 """
 
-SWITCH_ENFORCE_JWT_SCOPES = 'oauth2.enforce_jwt_scopes'
+NAMESPACE_SWITCH = 'oauth2'
+SWITCH_ENFORCE_JWT_SCOPES = 'enforce_jwt_scopes'
+NAMESPACED_SWITCH_ENFORCE_JWT_SCOPES = '{}.{}'.format(NAMESPACE_SWITCH, SWITCH_ENFORCE_JWT_SCOPES)
--- a/edx_rest_framework_extensions/jwt_decoder.py
+++ b/edx_rest_framework_extensions/jwt_decoder.py
@@ -2,14 +2,22 @@
 import logging
 
 import jwt
-import semantic_version
+from semantic_version import Version
 from django.conf import settings
 from rest_framework_jwt.settings import api_settings
 
 from edx_rest_framework_extensions.settings import get_jwt_issuers
 
-DEFAULT_JWT_SUPPORTED_VERSION = '1.0.0'
 logger = logging.getLogger(__name__)
+
+
+class JwtTokenVersion(object):
+    latest_supported = '1.1.0'
+
+    starting_version = '1.0.0'
+    added_version = '1.1.0'
+    added_is_restricted = '1.1.0'
+    added_filters = '1.1.0'
 
 
 def jwt_decode_handler(token):
@@ -71,7 +79,7 @@ def jwt_decode_handler(token):
 
     for jwt_issuer in get_jwt_issuers():
         try:
-            decoded = jwt.decode(
+            decoded_token = jwt.decode(
                 token,
                 jwt_issuer['SECRET_KEY'],
                 api_settings.JWT_VERIFY,
@@ -81,8 +89,9 @@ def jwt_decode_handler(token):
                 issuer=jwt_issuer['ISSUER'],
                 algorithms=[api_settings.JWT_ALGORITHM]
             )
-            verify_jwt_version(decoded)
-            return decoded
+            token_version = decode_and_verify_version(decoded_token)
+            upgraded_token = _upgrade_token(decoded_token, token_version)
+            return upgraded_token
         except jwt.InvalidTokenError:
             msg = "Token decode failed for issuer '{issuer}'".format(issuer=jwt_issuer['ISSUER'])
             logger.info(msg, exc_info=True)
@@ -92,19 +101,21 @@ def jwt_decode_handler(token):
     raise jwt.InvalidTokenError(msg)
 
 
-def verify_jwt_version(decoded_token):
+def decode_and_verify_version(decoded_token):
     """
     Verify that the JWT version is supported.
     """
-    supported_version = semantic_version.Version(
-        settings.JWT_AUTH.get('JWT_SUPPORTED_VERSION', DEFAULT_JWT_SUPPORTED_VERSION)
+    supported_version = Version(
+        settings.JWT_AUTH.get('JWT_SUPPORTED_VERSION', JwtTokenVersion.latest_supported)
     )
-    jwt_version = semantic_version.Version(
-        decoded_token.get('version', str(supported_version))
+    jwt_version = Version(
+        decoded_token.get('version', JwtTokenVersion.starting_version)
     )
     if jwt_version.major > supported_version.major:
         logger.info('Token decode failed due to unsupported JWT version number [%s]', str(jwt_version))
         raise jwt.InvalidTokenError
+
+    return jwt_version
 
 
 def decode_jwt_scopes(token):
@@ -114,9 +125,54 @@ def decode_jwt_scopes(token):
     return jwt_decode_handler(token).get('scopes', [])
 
 
+def decode_jwt_is_restricted(token):
+    """
+    Decode the JWT and return the is_restricted claim.
+    """
+    return jwt_decode_handler(token)['is_restricted']
+
+
 def decode_jwt_filters(token):
     """
-    Decode the JWT, parse the filters clain, and return a
-    list of (provider_type, filter_value) tuples.
+    Decode the JWT, parse the filters claim, and return a
+    list of (filter_type, filter_value) tuples.
     """
-    return [jwt_filter.split(':') for jwt_filter in jwt_decode_handler(token).get('filters', [])]
+    return [jwt_filter.split(':') for jwt_filter in jwt_decode_handler(token)['filters']]
+
+
+def _upgrade_token(token, token_version):
+    """
+    Returns an updated token that includes default values for
+    fields that were introduced since the token was created
+    by checking its version number.
+    """
+    def _upgrade_version(token, token_version):
+        """
+        Tokens didn't always contain a version number so we
+        default to a nominal starting number.
+        """
+        token['version'] = token.get('version', str(token_version))
+
+    def _upgrade_is_restricted(token, token_version):
+        """
+        We can safely default to False since all "restricted" tokens
+        created prior to this version were always created as expired
+        tokens. Expired tokens would not validate and so would
+        not get as this far into the decoding process.
+        """
+        if token_version < Version(JwtTokenVersion.added_is_restricted):
+            token['is_restricted'] = token.get('is_restricted', False)
+
+    def _upgrade_filters(token, token_version):
+        """
+        We can safely default to an empty list of filters since
+        previously created tokens were either "restricted" (always 
+        expired) or had full access.
+        """
+        if token_version < Version(JwtTokenVersion.added_filters):
+            token['filters'] = token.get('filters', [])
+
+    _upgrade_version(token, token_version)
+    _upgrade_is_restricted(token, token_version)
+    _upgrade_filters(token, token_version)
+    return token
--- a/edx_rest_framework_extensions/permissions.py
+++ b/edx_rest_framework_extensions/permissions.py
@@ -1,12 +1,18 @@
 """ Permission classes. """
+import logging
 from django.core.exceptions import ImproperlyConfigured
 from opaque_keys.edx.keys import CourseKey
-from rest_framework.permissions import BasePermission
+from rest_condition import C
+from rest_framework.authentication import SessionAuthentication
+from rest_framework.permissions import BasePermission, IsAuthenticated
 import waffle
 
 from edx_rest_framework_extensions.authentication import is_jwt_authenticated
-from edx_rest_framework_extensions.config import SWITCH_ENFORCE_JWT_SCOPES
-from edx_rest_framework_extensions.jwt_decoder import decode_jwt_filters, decode_jwt_scopes
+from edx_rest_framework_extensions.config import NAMESPACED_SWITCH_ENFORCE_JWT_SCOPES
+from edx_rest_framework_extensions.jwt_decoder import decode_jwt_filters, decode_jwt_scopes, decode_jwt_is_restricted
+
+
+log = logging.getLogger(__name__)
 
 
 class IsSuperuser(BasePermission):
@@ -29,20 +35,33 @@ class IsUserInUrl(BasePermission):
     Allows access if the requesting user matches the user in the URL.
     """
     def has_permission(self, request, view):
-        user_parameter_name = 'username'
-        url_username = (
-            request.GET.get(user_parameter_name, '') or
-            getattr(request, 'parser_context', {}).get('kwargs', {}).get(user_parameter_name, '')
-        )
-        return request.user.username.lower() == url_username.lower()
+        allowed = request.user.username.lower() == _get_username_param(request)
+        if not allowed:
+            log.info(u"Permission IsUserInUrl: not satisfied for requesting user %s.", request.user.username)
+        return allowed
 
 
-class IsJwtAuthenticated(BasePermission):
+class JwtRestrictedApplication(BasePermission):
     """
-    Returns whether the request was successfully authenticated with JwtAuthentication.
+    Returns whether the request was successfully authenticated with JwtAuthentication
+    by a RestrictedApplication.
     """
+    message = 'Not a Restricted JWT Application.'
+
     def has_permission(self, request, view):
-        return is_jwt_authenticated(request)
+        return self.is_enforced_and_jwt_restricted_app(request)
+
+    @classmethod
+    def is_enforced_and_jwt_restricted_app(cls, request):
+        is_enforcement_enabled = waffle.switch_is_active(NAMESPACED_SWITCH_ENFORCE_JWT_SCOPES)
+        ret_val = is_enforcement_enabled and is_jwt_authenticated(request) and decode_jwt_is_restricted(request.auth)
+        log.debug(u"Permission JwtRestrictedApplication: returns %s.", ret_val)
+        return ret_val
+
+
+class NotJwtRestrictedApplication(BasePermission):
+    def has_permission(self, request, view):
+        return not JwtRestrictedApplication.is_enforced_and_jwt_restricted_app(request)
 
 
 class JwtHasScope(BasePermission):
@@ -52,22 +71,16 @@ class JwtHasScope(BasePermission):
     message = 'JWT missing required scopes.'
 
     def has_permission(self, request, view):
-        if _should_skip_jwt_scope_enforcement(request):
-            return True
         jwt_scopes = decode_jwt_scopes(request.auth)
-        required_scopes = set(self.get_scopes(request, view))
-        if required_scopes.issubset(jwt_scopes):
-            return True
-        return False
-
-    def get_scopes(self, request, view):
-        """
-        Return the required scopes defined on the view.
-        """
-        try:
-            return getattr(view, 'required_scopes')
-        except AttributeError:
-            raise ImproperlyConfigured('JwtHasScope requires the view to define the required_scopes attribute')
+        required_scopes = set(getattr(view, 'required_scopes', []))
+        allowed = bool(required_scopes) and required_scopes.issubset(jwt_scopes)
+        if not allowed:
+            log.warning(
+                u"Permission JwtHasScope: required scopes '%s' are not a subset of the token's scopes '%s'.",
+                required_scopes,
+                jwt_scopes,
+            )
+        return allowed
 
 
 class JwtHasContentOrgFilterForRequestedCourse(BasePermission):
@@ -83,21 +96,69 @@ class JwtHasContentOrgFilterForRequestedCourse(BasePermission):
         of the organizations specified in the content provider filters
         in the JWT used to authenticate.
         """
-        if _should_skip_jwt_scope_enforcement(request):
-            return True
         course_key = CourseKey.from_string(view.kwargs.get('course_id'))
         jwt_filters = decode_jwt_filters(request.auth)
-        for provider_type, filter_value in jwt_filters:
-            if provider_type == 'content_org' and filter_value == course_key.org:
+        for filter_type, filter_value in jwt_filters:
+            if filter_type == 'content_org' and filter_value == course_key.org:
                 return True
+        log.warning(
+            u"Permission JwtHasContentOrgFilterForRequestedCourse: no filter found for %s.",
+            course_key.org,
+        )
         return False
 
 
-def _should_skip_jwt_scope_enforcement(request):
+class JwtHasUserFilterForRequestedUser(BasePermission):
     """
-    Returns True if either the request is not JWT authenticated or if
-    JWT scopes enforcement is disabled.
+    The JWT used to authenticate contains the appropriate user filter for the
+    requested user resource.
     """
-    if not is_jwt_authenticated(request):
-        return True
-    return not waffle.switch_is_active(SWITCH_ENFORCE_JWT_SCOPES)
+    message = 'JWT missing required user filter.'
+
+    def has_permission(self, request, view):
+        """
+        If the JWT has a user filter, verify that the filtered
+        user value matches the user in the URL.
+        """
+        user_filter = self._get_user_filter(request)
+        if not user_filter:
+            # no user filters are present in the token to limit access
+            return True
+
+        username_param = _get_username_param(request)
+        allowed = user_filter == username_param
+        if not allowed:
+            log.warning(
+                u"Permission JwtHasUserFilterForRequestedUser: user_filter %s doesn't match username %s.",
+                user_filter,
+                username_param,
+            )
+        return allowed
+
+    def _get_user_filter(self, request):
+        jwt_filters = decode_jwt_filters(request.auth)
+        for filter_type, filter_value in jwt_filters:
+            if filter_type == 'user':
+                if filter_value == 'me':
+                    filter_value = request.user.username.lower()
+                return filter_value
+        return None
+
+
+_NOT_JWT_RESTRICTED_PERMISSIONS = C(NotJwtRestrictedApplication) & (C(IsStaff) | IsUserInUrl)
+_JWT_RESTRICTED_PERMISSIONS = (
+    C(JwtRestrictedApplication) &
+    JwtHasScope &
+    JwtHasContentOrgFilterForRequestedCourse &
+    JwtHasUserFilterForRequestedUser
+)
+JWT_RESTRICTED_APPLICATION_OR_USER_ACCESS = C(IsAuthenticated) & (_NOT_JWT_RESTRICTED_PERMISSIONS | _JWT_RESTRICTED_PERMISSIONS)
+
+
+def _get_username_param(request):
+    user_parameter_name = 'username'
+    url_username = (
+        getattr(request, 'parser_context', {}).get('kwargs', {}).get(user_parameter_name, '') or
+        request.GET.get(user_parameter_name, '')
+    )
+    return url_username.lower()
--- a/edx_rest_framework_extensions/tests/test_jwt_decoder.py
+++ b/edx_rest_framework_extensions/tests/test_jwt_decoder.py
@@ -12,15 +12,15 @@ from edx_rest_framework_extensions.tests.factories import UserFactory
 from edx_rest_framework_extensions.jwt_decoder import jwt_decode_handler
 
 
-def generate_jwt(user, scopes=None, filters=None):
+def generate_jwt(user, scopes=None, filters=None, is_restricted=None):
     """
     Generate a valid JWT for authenticated requests.
     """
-    access_token = generate_jwt_payload(user, scopes=scopes, filters=filters)
-    return generate_jwt_token(access_token)
+    access_token = _generate_latest_version_payload(user, scopes=scopes, filters=filters, is_restricted=is_restricted)
+    return _generate_jwt_token(access_token)
 
 
-def generate_jwt_token(payload, signing_key=None):
+def _generate_jwt_token(payload, signing_key=None):
     """
     Generate a valid JWT token for authenticated requests.
     """
@@ -28,13 +28,32 @@ def generate_jwt_token(payload, signing_key=None):
     return jwt.encode(payload, signing_key).decode('utf-8')
 
 
-def generate_jwt_payload(user, scopes=None, filters=None, version='1.0.0'):
+def _generate_latest_version_payload(user, scopes=None, filters=None, version=None, is_restricted=None):
     """
     Generate a valid JWT payload given a user and optionally scopes and filters.
     """
+    payload = _generate_starting_version_payload(user)
+    payload.update({
+        # fix this version and add newly introduced fields as the version updates.
+        'version': '1.1.0',
+        'filters': [],
+        'is_restricted': False,
+    })
+    if scopes is not None:
+        payload['scopes'] = scopes
+    if version is not None:
+        payload['version'] = version
+    if filters is not None:
+        payload['filters'] = filters
+    if is_restricted is not None:
+        payload['is_restricted'] = is_restricted
+    return payload
+
+
+def _generate_starting_version_payload(user):
     jwt_issuer_data = settings.JWT_AUTH['JWT_ISSUERS'][0]
     now = int(time())
-    ttl = 5
+    ttl = 600
     payload = {
         'iss': jwt_issuer_data['ISSUER'],
         'aud': jwt_issuer_data['AUDIENCE'],
@@ -42,13 +61,8 @@ def generate_jwt_payload(user, scopes=None, filters=None, version='1.0.0'):
         'email': user.email,
         'iat': now,
         'exp': now + ttl,
+        'scopes': [],
     }
-    if version:
-        payload['version'] = version
-    if scopes:
-        payload['scopes'] = scopes
-    if filters:
-        payload['filters'] = filters
     return payload
 
 
@@ -76,17 +90,17 @@ class JWTDecodeHandlerTests(TestCase):
     def setUp(self):
         super(JWTDecodeHandlerTests, self).setUp()
         self.user = UserFactory()
-        self.payload = generate_jwt_payload(self.user)
-        self.jwt = generate_jwt_token(self.payload)
+        self.payload = _generate_latest_version_payload(self.user)
+        self.jwt = _generate_jwt_token(self.payload)
 
-    def test_decode_success(self):
+    def test_success(self):
         """
         Confirms that the format of the valid response from the token decoder matches the payload
         """
         self.assertDictEqual(jwt_decode_handler(self.jwt), self.payload)
 
     @ddt.data(*settings.JWT_AUTH['JWT_ISSUERS'])
-    def test_decode_valid_token_multiple_valid_issuers(self, jwt_issuer):
+    def test_valid_token_multiple_valid_issuers(self, jwt_issuer):
         """
         Validates that a valid token is properly decoded given a list of multiple valid issuers
         """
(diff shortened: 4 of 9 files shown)


edx_rest_framework_extensions/permissions.py
```diff
@@ -83,21 +90,61 @@ def has_permission(self, request, view):
...
+        JwtHasUserFilterForRequestedUser
+    )
+    return (not_jwt_restricted_permissions | jwt_restricted_permissions)
+
+
+def _get_username_param(request):
+    user_parameter_name = 'username'
+    url_username = (
+        request.GET.get(user_parameter_name, '') or
+        getattr(request, 'parser_context', {}).get('kwargs', {}).get(user_parameter_name, '')
```
ormsbee: Since `or` short-circuits, do we have to worry about people appending a different `username` to querystrings in addition to a `username` that comes up in a URL? Like `/users/ormsbee/books?username=someoneelse`

edx_rest_framework_extensions/tests/test_jwt_decoder.py
```diff
@@ -49,6 +49,8 @@ def generate_jwt_payload(user, scopes=None, filters=None, version='1.0.0'):
         payload['scopes'] = scopes
     if filters:
         payload['filters'] = filters
+    if is_restricted is not None:
+        payload['is_restricted'] = is_restricted
```
ormsbee: Why is `is_restricted` not just boolean?

edx_rest_framework_extensions/permissions.py
```diff
@@ -83,21 +90,61 @@ def has_permission(self, request, view):
...
+
+def jwt_course_access_or_is_user_in_url_or_staff():
+    not_jwt_restricted_permissions = C(NotJwtRestrictedApplication) & (C(IsStaff) | IsUserInUrl)
+    jwt_restricted_permissions = (
+        C(JwtRestrictedApplication) &
+        JwtHasScope &
+        JwtHasContentOrgFilterForRequestedCourse &
+        JwtHasUserFilterForRequestedUser
+    )
+    return (not_jwt_restricted_permissions | jwt_restricted_permissions)
```
ormsbee: 1. Why have this be a function instead of another importable conditional permission?
2. Is there something else that we could name it that better expresses the intent of putting all these permissions together? Or maybe break it up a bit more so that the jwt_restricted_permissions are one condition and we put them together in the view/view-decorator code? I think it gets confusing when there are two "or" in the function name and you don't know where the parens are supposed to be.

edx_rest_framework_extensions/tests/test_jwt_decoder.py
nasthagiri: Since we didn't always include the `is_restricted` boolean, the `None` value allows it to be conditionally included.

edx_rest_framework_extensions/permissions.py
nasthagiri: 1. It would be good to have it be a conditional permission.  But the middleware check doesn't find the required permission if I include it within a new permission class.  Would it work if I inherit from `Conditional` directly?
2. Oh yes PLEASE.  I named this late last night.  I beg for a better name than this.

edx_rest_framework_extensions/permissions.py
nasthagiri: Do you think a single endpoint would support both ways?
The current implementation of `IsUserInUrl` in edx-platform also shortcuts, by the way.

ormsbee: I don't think a single endpoint would support both by design, but I'm worried it might get exploited. Say someone designed the API to read the username from the path, but then leaves the permissions check to be done with `IsUserInUrl`. Then someone crafts a URL like `/users/intended_user/books?username=attacker`. Then would it be possible for a user who's logged in as `attacker` to pass the permissions check for `intended_user` and get access to that resource (because the view code grabs `username` from the path but the permission check looks at the querystring first)?

edx_rest_framework_extensions/permissions.py
ormsbee: 1. Weird. I thought I fixed the middleware check to drill into all the conditional permissions and pass if the target permission class is in at least one of them. Could you write out the `permission_classes` value that the middleware isn't searching through correctly? I can try debugging it on my side.

2. Yeah, that is a tough one. Is this the sort of standard set of permissions we're expecting to put on a legacy endpoint for compatibility? Are there other common variations that will be used?

edx_rest_framework_extensions/config.py
```diff
@@ -2,4 +2,6 @@
 Application configuration constants and code.
 """
 
-SWITCH_ENFORCE_JWT_SCOPES = 'oauth2.enforce_jwt_scopes'
+NAMESPACE_SWITCH = 'oauth2'
```
robrap: I think the before code is simpler.  Not sure why this is needed.

edx_rest_framework_extensions/permissions.py
```diff
@@ -52,22 +69,12 @@ class JwtHasScope(BasePermission):
...
-        Return the required scopes defined on the view.
-        """
-        try:
-            return getattr(view, 'required_scopes')
-        except AttributeError:
-            raise ImproperlyConfigured('JwtHasScope requires the view to define the required_scopes attribute')
+        required_scopes = set(getattr(view, 'required_scopes', []))
+        allowed = bool(required_scopes) and required_scopes.issubset(jwt_scopes)
+        if not allowed:
+            log.warning("Permission JwtHasScope not satisfied.")
```
robrap: Output of required_scopes and jwt_scopes would be useful, if not computing the missing scope.  Are there other details from the request that one would want if they were looking at these warnings trying to learn something?

edx_rest_framework_extensions/permissions.py
```diff
@@ -83,21 +90,61 @@ def has_permission(self, request, view):
...
-        if _should_skip_jwt_scope_enforcement(request):
-            return True
         course_key = CourseKey.from_string(view.kwargs.get('course_id'))
         jwt_filters = decode_jwt_filters(request.auth)
-        for provider_type, filter_value in jwt_filters:
-            if provider_type == 'content_org' and filter_value == course_key.org:
+        for filter_type, filter_value in jwt_filters:
+            if filter_type == 'content_org' and filter_value == course_key.org:
                 return True
+        log.warning("Permission JwtHasContentOrgFilterForRequestedCourse not satisfied.")
```
robrap: Same logging detail question.  Would it be useful to know about details of an org mismatch for example?

edx_rest_framework_extensions/jwt_decoder.py
```diff
@@ -114,9 +114,16 @@ def decode_jwt_scopes(token):
     return jwt_decode_handler(token).get('scopes', [])
 
 
+def decode_jwt_is_restricted(token):
+    """
+    Decode the JWT and return the is_restricted claim.
+    """
+    return jwt_decode_handler(token).get('is_restricted', True)
```
douglashall: Will defaulting to `True` be a problem for existing tokens out in the wild that do not have the `is_restricted` claim?

edx_rest_framework_extensions/permissions.py
```diff
@@ -83,21 +90,61 @@ def has_permission(self, request, view):
...
+        If the JWT has a user filter, verify that the filtered
+        user value matches the user in the URL.
+        """
+        user_filter = self._get_user_filter(request)
+        if not user_filter:
+            return True
+
+        allowed = user_filter == _get_username_param(request)
+        if not allowed:
+            log.warning("Permission JwtHasUserFilterForRequestedUser not satisfied for user_filter %s.", user_filter)
```
robrap: Again, would we want `_get_username_param(request)`?

edx_rest_framework_extensions/jwt_decoder.py
```diff
@@ -114,9 +114,16 @@ def decode_jwt_scopes(token):
...
+
+
 def decode_jwt_filters(token):
     """
     Decode the JWT, parse the filters clain, and return a
```
douglashall: I made a typo previously, s/clain/claim/.

edx_rest_framework_extensions/permissions.py
```diff
@@ -83,21 +90,61 @@ def has_permission(self, request, view):
...
+        return allowed
+
+    def _get_user_filter(self, request):
+        jwt_filters = decode_jwt_filters(request.auth)
+        for filter_type, filter_value in jwt_filters:
+            if filter_type == 'user':
+                if filter_value == 'me':
```
robrap: Do we really need to support "me" in a way that could conflict with a "me" username?

edx_rest_framework_extensions/permissions.py
robrap: Whatever is decided, I think you should capture this in a unit test.

edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -13,7 +13,12 @@
 from rest_framework.views import APIView
 
 from ..middleware import EnsureJWTAuthSettingsMiddleware
-from ..permissions import IsJwtAuthenticated, IsSuperuser, IsStaff, JwtHasScope
+from ..permissions import (
+    IsSuperuser,
+    IsStaff,
```
robrap: Guessing we aren't using isort.

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -22,15 +23,13 @@ class IsSuperuserTests(TestCase):
 
     @ddt.data(True, False)
     def test_has_permission(self, has_permission):
```
robrap: Rename to `test_superuser_has_permission`.

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -64,19 +63,39 @@ def test_has_permission(self, has_user_in_url):
 
 
 @ddt.ddt
-class IsJwtAuthenticatedTests(TestCase):
-    """ Tests for the IsJwtAuthenticated permission class. """
+class JwtRestrictedApplicationTests(TestCase):
```
robrap: Is `JwtApplicationTests` too meaningless?

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -64,19 +63,39 @@ def test_has_permission(self, has_user_in_url):
...
-        self.assertEqual(
-            permissions.IsJwtAuthenticated().has_permission(request, None),
-            issubclass(type(request.successful_authenticator), BaseJSONWebTokenAuthentication),
-        )
+        request.auth = generate_jwt(request.user, is_restricted=is_restricted)
+
+        is_jwt_auth_subclass = issubclass(type(request.successful_authenticator), BaseJSONWebTokenAuthentication)
+
+        has_permission = permission_class().has_permission(request, view=None)
+        if permission_class == permissions.JwtRestrictedApplication:
```
robrap: If you used:
```
expected_restricted_permission = enforce_scopes and is_restricted and is_jwt_auth_subclass
```
Below, you could use `expected_restricted_permission` and `not expected_restricted_permission`.

edx_rest_framework_extensions/jwt_decoder.py
nasthagiri: You're right.  In fact, this is a great time where I can use the `version` number on the token to make a safe assumption of whether or not to expect `is_restricted`.  😄

edx_rest_framework_extensions/jwt_decoder.py
nasthagiri: No worries. I got your back.  Fixed.  😄

edx_rest_framework_extensions/config.py
nasthagiri: The edx-platform imports the switch from here and needs the namespace value to use waffle_utils.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: done.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: I renamed it to `JwtApplicationPermissionsTests`.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: True. Updated.

edx_rest_framework_extensions/permissions.py
nasthagiri: Done.

edx_rest_framework_extensions/permissions.py
nasthagiri: done.

edx_rest_framework_extensions/permissions.py
nasthagiri: I added the org name in the warning message.

edx_rest_framework_extensions/permissions.py
nasthagiri: Good catch.  I reversed the clauses in the `or` statement and added a unit test.

edx_rest_framework_extensions/permissions.py
nasthagiri: 1. The middleware does work as designed and drills in while the Permission class is a Conditional.  But, if I wanted to do the following:

```
class CourseUserAccessPermission(BasePermission):
    def has_permission(self, request, view):
        not_jwt_restricted_permissions = C(NotJwtRestrictedApplication) & (C(IsStaff) | IsUserInUrl)
        jwt_restricted_permissions = (
            C(JwtRestrictedApplication) &
            JwtHasScope &
            JwtHasContentOrgFilterForRequestedCourse &
            JwtHasUserFilterForRequestedUser
        )
        return (not_jwt_restricted_permissions | jwt_restricted_permissions)
```

then it wouldn't catch that the implementation of the `Permission` class checks the required permission within it.

nasthagiri: 2. What about:
```
jwt_restricted_application_or_user_access
```

edx_rest_framework_extensions/tests/test_jwt_decoder.py
```diff
@@ -178,7 +189,7 @@ def test_decode_unsupported_jwt_version(self):
 
             # Verify that the proper entries were written to the log file
             msg = "Token decode failed due to unsupported JWT version number [%s]"
-            patched_log.info.assert_any_call(msg, '1.0.0')
+            patched_log.info.assert_any_call(msg, '1.1.0')
```
robrap: Why is this 1.1.0?

edx_rest_framework_extensions/tests/test_jwt_decoder.py
```diff
@@ -149,24 +160,24 @@ def test_decode_failure_invalid_token(self):
...
         """
         token = generate_jwt_token(self.payload)
         self.assertDictEqual(jwt_decode_handler(token), self.payload)
 
     @ddt.data(None, '0.5.0', '1.0.0', '1.0.5', '1.5.0', '1.5.5')
-    def test_decode_supported_jwt_version(self, jwt_version):
+    def test_supported_jwt_version(self, jwt_version):
         """
-        Verifies the JWT is decoded successfully when the JWT_SUPPORTED_VERSION setting is not specified.
+        Verifies the JWT is decoded successfully with different supported versions in the token.
```
robrap: What is the point of this test?  Why do we need so many values in the ddt (more test time)?  I don't think this is testing for "supported versions", but just versions?  Do we want to do any sort of assertion with `jwt_token`?

edx_rest_framework_extensions/tests/test_jwt_decoder.py
```diff
@@ -149,24 +160,24 @@ def test_decode_failure_invalid_token(self):
...
         """
         jwt_payload = generate_jwt_payload(self.user, version=jwt_version)
         token = generate_jwt_token(jwt_payload)
         self.assertDictEqual(jwt_decode_handler(token), jwt_payload)
 
     @override_settings(JWT_AUTH=update_jwt_auth_setting({'JWT_SUPPORTED_VERSION': '0.5.0'}))
-    def test_decode_unsupported_jwt_version(self):
+    def test_unsupported_jwt_version(self):
```
robrap: This test is doubling as both a test of unsupported version and a test for overriding version using the setting.  Can the name be updated, or can the docstring mention this?

edx_rest_framework_extensions/tests/test_jwt_decoder.py
```diff
@@ -188,3 +199,17 @@ def test_decode_unsupported_jwt_version(self):
...
+        Verifies the JWT is upgraded when an old (starting) version is provided.
+
+        Compares the decoded token with the latest payload version that would
+        other be generated, while keeping time-related values constant.
+        """
+        jwt_payload = generate_starting_version(self.user)
+        token = generate_jwt_token(jwt_payload)
+
+        upgraded_payload = generate_jwt_payload(self.user, version='1.0.0')
+        upgraded_payload['iat'], upgraded_payload['exp'] = jwt_payload['iat'], jwt_payload['exp']
```
robrap: Are these the time-related values?  I didn't understand this line at all.  Can we remove the comment from the docstring and instead have this line commented with something like:
```
Keep time-related values constant for easier comparison.
```

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -51,19 +51,53 @@ def test_has_permission(self, is_staff):
...
+            user,
+            username_in_resource=user.username if has_user_in_resource else 'another_user',
+        )
+        self.assertEqual(permissions.IsUserInUrl().has_permission(request, None), has_user_in_resource)
+
+    def test_resource_takes_precedence_over_param(self):
+        user = factories.UserFactory(username='this_user')
+        request = self._create_request(
+            user,
+            username_in_resource='another_user',
```
robrap: Nit: Could use ddt with username_in_resource, username_in_param, and expected_has_permission, but fine as-is as well.

edx_rest_framework_extensions/permissions.py
```diff
@@ -142,15 +144,14 @@ def _get_user_filter(self, request):
...
-def jwt_course_access_or_is_user_in_url_or_staff():
-    not_jwt_restricted_permissions = C(NotJwtRestrictedApplication) & (C(IsStaff) | IsUserInUrl)
-    jwt_restricted_permissions = (
-        C(JwtRestrictedApplication) &
-        JwtHasScope &
-        JwtHasContentOrgFilterForRequestedCourse &
-        JwtHasUserFilterForRequestedUser
-    )
-    return (not_jwt_restricted_permissions | jwt_restricted_permissions)
+NOT_JWT_RESTRICTED_PERMISSIONS = C(NotJwtRestrictedApplication) & (C(IsStaff) | IsUserInUrl)
```
robrap: Are these meant to be used outside of `JWT_RESTRICTED_APPLICATION_OR_USER_ACCESS`?
If no, maybe add `_` prefix?
If yes, should IsAuthenticated condition be included in these conditions?  Might want to do this either way.

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -184,3 +185,114 @@ def test_has_permission(self, authentication_class, jwt_filters, view_kwargs, ex
...
+    @patch('edx_rest_framework_extensions.permissions.waffle.switch_is_active')
+    def test_jwt_unrestricted_succeeds(self, waffle_mock):
+        waffle_mock.return_value = True
+        request = self._create_request(username_in_url=True)
+        self._create_jwt(request)
+
+        response = self.SomeClassView().dispatch(request)
+        self.assertEqual(response.status_code, 200)
+
+    # @patch('edx_rest_framework_extensions.permissions.waffle.switch_is_active')
```
robrap: Reminder to do whatever you intended to this. :)

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -184,3 +185,114 @@ def test_has_permission(self, authentication_class, jwt_filters, view_kwargs, ex
...
+        self._create_session(request)
+
+        response = self.SomeClassView().dispatch(request)
+        self.assertEqual(response.status_code, 200)
+
+    @patch('edx_rest_framework_extensions.permissions.waffle.switch_is_active')
+    def test_jwt_unrestricted_succeeds(self, waffle_mock):
+        waffle_mock.return_value = True
+        request = self._create_request(username_in_url=True)
+        self._create_jwt(request)
```
robrap: If the jwt included is_restricted=True would this fail?  Should it?
Two separate questions related to my uncertainty around restricted. :)

edx_rest_framework_extensions/tests/test_jwt_decoder.py
```diff
@@ -28,29 +28,40 @@ def generate_jwt_token(payload, signing_key=None):
     return jwt.encode(payload, signing_key).decode('utf-8')
 
 
-def generate_jwt_payload(user, scopes=None, filters=None, version='1.0.0', is_restricted=None):
+def generate_jwt_payload(user, scopes=None, filters=None, version=None, is_restricted=None):
     """
     Generate a valid JWT payload given a user and optionally scopes and filters.
     """
+    payload = generate_starting_version(user)
+    payload.update({
```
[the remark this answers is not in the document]
nasthagiri: The tests use this helper to verify that the resulting JWT matches the expected value.  Therefore, the tests must get the latest version here.
I renamed this method to `_generate_latest_version_payload` and added a comment to update the fields and the version here whenever the version updates.  Tests will automatically fail when the version is updated.

edx_rest_framework_extensions/tests/test_jwt_decoder.py
```diff
@@ -28,29 +28,40 @@ def generate_jwt_token(payload, signing_key=None):
...
+        'version': '1.1.0',
```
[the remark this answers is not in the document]
nasthagiri: The constant is intentionally not used here since we want people to manually update this and the expected payload when the version number changes.  I've added a comment to this effect.

edx_rest_framework_extensions/tests/test_jwt_decoder.py
```diff
@@ -28,29 +28,40 @@ def generate_jwt_token(payload, signing_key=None):
...
+    if version is not None:
+        payload['version'] = version
+    if filters is not None:
+        payload['filters'] = filters
+    if is_restricted is not None:
+        payload['is_restricted'] = is_restricted
+    return payload
+
+
+def generate_starting_version(user):
```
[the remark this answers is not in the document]
nasthagiri: Renamed to `_generate_starting_version_payload`.

edx_rest_framework_extensions/tests/test_jwt_decoder.py
```diff
@@ -99,7 +110,7 @@ def test_decode_valid_token_multiple_valid_issuers(self, jwt_issuer):
         token = generate_jwt_token(self.payload, jwt_issuer['SECRET_KEY'])
         self.assertEqual(jwt_decode_handler(token), self.payload)
 
-    def test_decode_failure(self):
+    def test_failure(self):
```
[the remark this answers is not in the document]
nasthagiri: done

edx_rest_framework_extensions/tests/test_jwt_decoder.py
nasthagiri: done.

edx_rest_framework_extensions/tests/test_jwt_decoder.py
nasthagiri: I'm not addressing these concerns with the previously existing tests for now.  We can revisit this once Hawthorn is unblocked.

edx_rest_framework_extensions/tests/test_jwt_decoder.py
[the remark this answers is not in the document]
robrap: Might want an assertion to test this hard-coded value against the constant to force the test to break, rather than relying on a comment.

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -221,78 +223,140 @@ def test_has_permission(self, user_filters, requested_username, expected_result)
...
+        # **** Enforced ****
+        # unrestricted
+        JwtTestCase(is_enforced=True, is_restricted=False, is_user_in_url=True, expected_response=200, expected_log=JwtIsUnrestrictedLog),
+        JwtTestCase(is_enforced=True, is_restricted=False, is_user_in_url=False, expected_response=403, expected_log=IsUserInUrlLog),
+
+        # restricted
+        JwtTestCase(is_enforced=True, is_restricted=True, is_user_in_url=True, expected_response=403, expected_log=JwtHasScopeLog),
+        JwtTestCase(is_enforced=True, is_restricted=True, is_user_in_url=False, expected_response=403, expected_log=JwtHasScopeLog),
+
+        # **** Unenforced ****
```
robrap: Nit: Easier on the mind if Unenforced came first, to match the order of unrestricted/restricted.

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -155,3 +182,176 @@ def test_has_permission(self, enforce_scopes, authentication_class, jwt_filters,
...
+
+
+@ddt.ddt
+class JwtRestrictionApplicationOrUserAccessTests(TestCase):
+    ExpectedLog = namedtuple('ExpectedLog', 'method, text, param')
+
+    IsUserInUrlLog = ExpectedLog('info', 'IsUserInUrl', None)
+    JwtIsUnrestrictedLog = ExpectedLog('debug', 'JwtRestrictedApplication', False)
+    JwtIsRestrictedLog = ExpectedLog('debug', 'JwtRestrictedApplication', True)
+    JwtHasScopeLog = ExpectedLog('warning', 'JwtHasScope', None)
```
robrap: Nit: might be more clear to name this something like `MissingScopeForJwtHasScopeLog`.  Other expected logs might be the same.  It would make the tests more clear why they are failing when this name is the only indication.

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -155,3 +182,176 @@ def test_has_permission(self, enforce_scopes, authentication_class, jwt_filters,
...
+        # unrestricted
+        JwtTestCase(is_enforced=False, is_restricted=False, is_user_in_url=True, expected_response=200, expected_log=JwtIsUnrestrictedLog),
+        JwtTestCase(is_enforced=False, is_restricted=False, is_user_in_url=False, expected_response=403, expected_log=IsUserInUrlLog),
+
+        # restricted
+        JwtTestCase(is_enforced=False, is_restricted=True, is_user_in_url=True, expected_response=200, expected_log=JwtIsUnrestrictedLog),
+        JwtTestCase(is_enforced=False, is_restricted=True, is_user_in_url=False, expected_response=403, expected_log=IsUserInUrlLog),
+    )
+    @ddt.unpack
+    def test_jwt_cases(self, is_enforced, is_restricted, is_user_in_url, expected_response, expected_log, mock_log, waffle_mock):
```
robrap: Is this `test_jwt_cases_without_scopes_and_filters`? Other?  I think better names would make it more clear if we were missing coverage.

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -155,3 +182,176 @@ def test_has_permission(self, enforce_scopes, authentication_class, jwt_filters,
...
+        JwtRestrictedTestCase(scopes=['required_scope'], filters=['content_org:some_org', 'some:other'], expected_response=200, expected_log=JwtIsRestrictedLog),
+
+        JwtRestrictedTestCase(scopes=['required_scope'], filters=['content_org:another_org'], expected_response=403, expected_log=JwtOrgFilterLog),
+        JwtRestrictedTestCase(scopes=['required_scope'], filters=[], expected_response=403, expected_log=JwtOrgFilterLog),
+
+        JwtRestrictedTestCase(scopes=[], filters=['content_org:some_org'], expected_response=403, expected_log=JwtHasScopeLog),
+        JwtRestrictedTestCase(scopes=['another_scope'], filters=['content_org:some_org'], expected_response=403, expected_log=JwtHasScopeLog),
+    )
+    @ddt.unpack
+    def test_jwt_enforced_restricted(self, scopes, filters, expected_response, expected_log, mock_log, waffle_mock):
```
robrap: I thought this should be `test_jwt_with_scopes_and_filters`?  But, is it `test_jwt_org_filters`?  Do we have consistent coverage for user filter?  I don't think it is quite the same, because you didn't mess with scopes.  Do you need to mess with scopes in this test, or could required_scopes always be there and be removed from the params?

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -155,3 +182,176 @@ def test_has_permission(self, enforce_scopes, authentication_class, jwt_filters,
...
+        response = self.SomeClassView().dispatch(request)
+        self.assertEqual(response.status_code, expected_response)
+        self._assert_log(mock_log, expected_log)
+
+    JwtRestrictedTestCase = namedtuple('JwtRestrictedTestCase', 'scopes, filters, expected_response, expected_log')
+
+    @patch('edx_rest_framework_extensions.permissions.waffle.switch_is_active')
+    @patch('edx_rest_framework_extensions.permissions.log')
+    @ddt.data(
+        JwtRestrictedTestCase(scopes=['required_scope'], filters=['content_org:some_org'], expected_response=200, expected_log=JwtIsRestrictedLog),
```
robrap: Maybe comments?
```
# Valid org filters
...

# Invalid org filters
...

# Invalid scopes
...
```

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -155,3 +182,176 @@ def test_has_permission(self, enforce_scopes, authentication_class, jwt_filters,
...
+    def test_has_permission(self, user_filters, requested_username, expected_result):
+        request = self._create_request(user_filters, requested_username)
+        self.assertEqual(
+            permissions.JwtHasUserFilterForRequestedUser().has_permission(request, None),
+            expected_result,
+        )
+
+
+@ddt.ddt
+class JwtRestrictionApplicationOrUserAccessTests(TestCase):
```
robrap: ApplicationRestriction?

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: do this.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: do this.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: rename to something like 'ScopesError' maybe.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: do this.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: typo: change to JwtRestrictedApplicationOrUserAccessTests

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -155,3 +182,176 @@ def test_has_permission(self, enforce_scopes, authentication_class, jwt_filters,
...
+    @ddt.data(
+        JwtRestrictedTestCase(scopes=['required_scope'], filters=['content_org:some_org'], expected_response=200, expected_log=JwtIsRestrictedLog),
+        JwtRestrictedTestCase(scopes=['required_scope', 'another_scope'], filters=['content_org:some_org'], expected_response=200, expected_log=JwtIsRestrictedLog),
+        JwtRestrictedTestCase(scopes=['required_scope'], filters=['content_org:some_org', 'some:other'], expected_response=200, expected_log=JwtIsRestrictedLog),
+
+        JwtRestrictedTestCase(scopes=['required_scope'], filters=['content_org:another_org'], expected_response=403, expected_log=JwtOrgFilterLog),
+        JwtRestrictedTestCase(scopes=['required_scope'], filters=[], expected_response=403, expected_log=JwtOrgFilterLog),
+
+        JwtRestrictedTestCase(scopes=[], filters=['content_org:some_org'], expected_response=403, expected_log=JwtHasScopeLog),
+        JwtRestrictedTestCase(scopes=['another_scope'], filters=['content_org:some_org'], expected_response=403, expected_log=JwtHasScopeLog),
```
nasthagiri: Note to self: Need to add tests for user filters.

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -221,78 +223,140 @@ def test_has_permission(self, user_filters, requested_username, expected_result)
...
     @patch('edx_rest_framework_extensions.permissions.waffle.switch_is_active')
-    def test_jwt_unrestricted_succeeds(self, waffle_mock):
-        waffle_mock.return_value = True
-        request = self._create_request(username_in_url=True)
-        self._create_jwt(request)
+    @patch('edx_rest_framework_extensions.permissions.log')
+    @ddt.data(
+        # **** Enforced ****
+        # unrestricted
+        JwtTestCase(is_enforced=True, is_restricted=False, is_user_in_url=True, expected_response=200, expected_log=JwtIsUnrestrictedLog),
```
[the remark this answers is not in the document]
nasthagiri: updated.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: updated.

edx_rest_framework_extensions/tests/test_permissions.py
```diff
@@ -221,78 +223,140 @@ def test_has_permission(self, user_filters, requested_username, expected_result)
...
+    JwtIsUnrestrictedLog = ExpectedLog('debug', 'JwtRestrictedApplication', False)
+    JwtIsRestrictedLog = ExpectedLog('debug', 'JwtRestrictedApplication', True)
+    JwtHasScopeLog = ExpectedLog('warning', 'JwtHasScope', None)
+    JwtOrgFilterLog = ExpectedLog('warning', 'JwtHasContentOrgFilterForRequestedCourse', None)
+
     class SomeClassView(APIView):
-        authentication_classes = (SessionAuthentication, JwtAuthentication)
+        authentication_classes = (JwtAuthentication, SessionAuthentication)
         permission_classes = (permissions.JWT_RESTRICTED_APPLICATION_OR_USER_ACCESS,)
+        required_scopes = ['required_scope']
```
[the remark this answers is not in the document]
nasthagiri: From Robert: we can ignore this.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: updated.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: done.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: done.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: no longer needed since we broke up into smaller test methods.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: done.

edx_rest_framework_extensions/tests/test_permissions.py
nasthagiri: done.

PR #42, opened 2018-08-06
Set appropriate defaults for `is_restricted` and `filters` jwt attributes regardless of version.

douglashall:


edx_rest_framework_extensions/jwt_decoder.py
```diff
@@ -19,7 +19,6 @@ class JwtTokenVersion(object):
 
     starting_version = '1.0.0'
     added_version = '1.1.0'
-    added_is_restricted = '1.1.0'
     added_filters = '1.1.0'
```
nasthagiri: Reminder to remove `added_filters` as well.

edx_rest_framework_extensions/jwt_decoder.py
```diff
@@ -128,7 +127,7 @@ def _set_filters(token, token_version):
 
     token_version = _get_and_set_version(token)
     _verify_version(token_version)
-    _set_is_restricted(token, token_version)
+    _set_is_restricted(token)
     _set_filters(token, token_version)
```
nasthagiri: Reminder to always set default for `filters` as well.

PR #47, opened 2018-09-19
Add jwt_refresh_cookie_name

nasthagiri:


robrap:
@nasthagiri: This has been rebased off the latest.  My questions remain open, even they though they are now marked as "outdated", because the files moved.

edx_rest_framework_extensions/__init__.py
```diff
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '1.8.0'  # pragma: no cover
+__version__ = '1.9.0'  # pragma: no cover
```
robrap: @nasthagiri: FYI: This is the pattern I follow for versioning: https://semver.org/
Only backward-compatible bug fixes should update the final digit.  Any new functionality (backward-compatible) should bump the middle number as I have done here.

PR #48, opened 2018-09-20
Refactor authentication related code.

robrap:
Before merge:
- [ ] squash
- [ ] adjust commit comment to mention ARCH-244


edx_rest_framework_extensions/middleware.py
```diff
@@ -1,76 +1,19 @@
...
-    def process_view(self, request, view_func, view_args, view_kwargs):  # pylint: disable=unused-argument
-        # Views as functions store the view's class in the 'view_class' attribute.
-        # Viewsets store the view's class in the 'cls' attribute.
-        view_class = getattr(
-            view_func,
-            'view_class',
-            getattr(view_func, 'cls', view_func),
-        )
+from rest_framework_jwt.authentication import (
+    BaseJSONWebTokenAuthentication,
```
robrap: @nasthagiri: This is probably backward compatibility overkill, but means less of a chance of something going wrong.  It was just used in one of our classes, but someone could have imported it.  Should I leave this or kill it?

edx_rest_framework_extensions/middleware.py
```diff
@@ -1,76 +1,19 @@
...
+)  # pylint: disable=unused-import
+from edx_rest_framework_extensions.auth.jwt.middleware import (
+    EnsureJWTAuthSettingsMiddleware,
+    JwtAuthCookieMiddleware,
+)  # pylint: disable=unused-import
 
-        view_authentication_classes = getattr(view_class, 'authentication_classes', tuple())
-        if self._includes_base_class(view_authentication_classes, BaseJSONWebTokenAuthentication):
-            self._add_missing_jwt_permission_classes(view_class)
+from .permissions import NotJwtRestrictedApplication   # pylint: disable=unused-import
```
robrap: @nasthagiri: Same question about leaving or killing.  I'd hope we didn't import from here, and that we'd be ok with cleaning up an IDA if it did?

edx_rest_framework_extensions/middleware.py
robrap: FYI: I'm leaning toward removing these, but want you to weigh-in.

edx_rest_framework_extensions/middleware.py
robrap: FYI: I'm leaning toward removing these, but want you to weigh-in.

edx_rest_framework_extensions/middleware.py
nasthagiri: Sure. Just leave it. We already have a TODO here - so we can remove all BWC code at once.

edx_rest_framework_extensions/middleware.py
nasthagiri: see above.

PR #49, opened 2018-09-24
Added SessionAuthenticationAllowInactiveUser

ahsan-ul-haq:
LEARNER-6423

### DIFF
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '1.9.0'  # pragma: no cover
+__version__ = '1.10.0'  # pragma: no cover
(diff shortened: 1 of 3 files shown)


ahsan-ul-haq:
@nasthagiri Is this what you are suggesting?

nasthagiri:
@ahsan-ul-haq Two questions:

- Is this code any different from what's currently in edx-platform?
- Can you create a sister PR that deletes this code in edx-platform and imports this authentication class from here?  You can do this in a backward-compatible way so that existing python imports don't need to be updated.

ahsan-ul-haq:
@nasthagiri No this code is not different from the code being used in edx-platform.
Yes I am going to make a sister PR with the commit here and will also create a sandbox for mobile testing. I am trying to confirm that this is what you intended to say when you asked to move SessionAuthenticationAllowInactiveUser in edx-drf-extensions and we are sure that this is the place where it should reside.

PR #51, opened 2018-10-08
Move o-auth2 authentication allow inactive user to edx drf extension

ahsan-ul-haq:
LEARNER-6512

### DIFF
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '1.10.0'  # pragma: no cover
+__version__ = '1.11.0'  # pragma: no cover
--- a/edx_rest_framework_extensions/auth/bearer/authentication.py
+++ b/edx_rest_framework_extensions/auth/bearer/authentication.py
@@ -17,7 +17,7 @@ class BearerAuthentication(BaseAuthentication):
     """
     Simple token based authentication.
 
-    This authentication class is useful for authenticating an OAuth2 access token against a remote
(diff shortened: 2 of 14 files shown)


nasthagiri:
@ahsan-ul-haq I apologize for this miscommunication.  If the mobile app's use of completion API needs `OAuth2AuthenticationAllowInactiveUser`, moving it to `edx-drf-extensions` is counter to the direction we are headed.  `edx-drf-extensions` is meant to be used across all IDAs.  However, `OAuth2Authentication` (and the like) cannot be used in any IDA other than the LMS - since it depends on the OAuth2 tables that should be accessible by only the LMS.

On the other hand, `JWTAuthentication` can be used across all IDAs since JWTs are stateless tokens and don't need to make any SQL table queries.

My initial confusion stemmed from whether mobile apps really need to make API calls to completion or can they just update their completions when accessing components via their web-views?  This was an implicit assumption made in the original ticket: https://github.com/edx/completion/pull/29

We have now gone down a possibly non-viable path.  At this point, I believe we either need to (1) update the mobile apps to use JWTs (if they need to call any of our non-LMS APIs) or (2) break abstractions and have completion depend on the LMS.  Are the mobile apps calling any non-LMS APIs today?  And if so, how are they doing so?

FYI @staubina

ahsan-ul-haq:
@nasthagiri Thanks for detailed clearance on this. 
As per my knowledge this is the first time mobile team needed to call any non-LMS APIs.

robrap:
Since the comment https://github.com/edx/edx-drf-extensions/pull/51#issuecomment-427903909 states that this is the wrong direction, and this PR includes DOP related changes which is long out of date, I am closing this PR.

PR #52, opened 2018-10-12
ARCH-255: Remove backward compatible auth imports.

robrap:
ARCH-255


robrap:
@nasthagiri: FYI: This is ready for review, even though I plan to get the IDA PRs lined up before merging this.

edx_rest_framework_extensions/auth/session/authentication.py
```diff
@@ -1,16 +1,6 @@
-"""
-Deprecated imports for authentication classes that moved.
+""" Session Authentication classes. """
```
robrap: Since I am working to remove the backward compatibility files and update all the IDA, I decided to move this authentication class as part of this work.
FYI: @ahsan-ul-haq @nasthagiri

edx_rest_framework_extensions/__init__.py
```diff
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '1.11.0'  # pragma: no cover
+__version__ = '2.0.0'  # pragma: no cover
```
robrap: FYI: I'm not precious about version numbers.  Using 2.0 to represent a backward incompatible change.

PR #53, opened 2018-10-29
Require clients to explicitly request JWT cookie authentication.

douglashall:
**Related PR:** https://github.com/edx/frontend-auth/pull/5

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/middleware.py
+++ b/edx_rest_framework_extensions/auth/jwt/middleware.py
@@ -15,6 +15,7 @@ from edx_rest_framework_extensions.auth.jwt.constants import JWT_DELIMITER
 from edx_rest_framework_extensions.permissions import NotJwtRestrictedApplication
 
 log = logging.getLogger(__name__)
+USE_JWT_COOKIE_HEADER = 'HTTP_USE_JWT_COOKIE'
 
 
 class EnsureJWTAuthSettingsMiddleware(object):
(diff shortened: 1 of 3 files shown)


edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -93,7 +94,8 @@ class JwtAuthCookieMiddleware(object):
 
     Also, sets the metric 'request_jwt_cookie' with one of the following values:
         'yes': Value when reconstitution is successful.
-        'no': Value when both cookies are missing and reconstitution is not possible.
+        'no': Value when jwt cookie authentication was not requested by the client.
```
robrap: `yes` and `no` aren't really opposites.   Should we use `success` and `not-requested`?

douglashall: Sounds good. I will update.

robrap:
1. The metric being updated was never added to this page: https://openedx.atlassian.net/wiki/spaces/PERF/pages/54362736/Custom+Metrics+in+New+Relic
Note: We might want the Confluence page to just point to code for the values so it is more resilient to change.
2. I imagine you are only going to deploy as necessary.  I can probably bump the version deployed when I reorder the middleware for the other IDAs.

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -112,17 +114,21 @@ def process_request(self, request):
...
         Reconstitute the full JWT and add a new cookie on the request object.
         """
+        use_jwt_cookie = request.META.get(USE_JWT_COOKIE_HEADER)
         header_payload_cookie = request.COOKIES.get(jwt_cookie_header_payload_name())
         signature_cookie = request.COOKIES.get(jwt_cookie_signature_name())
 
-        # Reconstitute JWT auth cookie if split cookies are available.
-        if header_payload_cookie and signature_cookie:
+        if not use_jwt_cookie:
+            metric_value = 'not-requested'
```
nasthagiri: Nit: rename the `use_jwt_cookie` to `use_jwt_cookie_requested` - it will then be consistent with the metric value and may be clearer.

nasthagiri:
@robrap I'll add these metrics to that page.  Thanks for the reminder.

PR #54, opened 2018-10-30
ARCH-262: Minor clean-up of authn.

robrap:
1. Add a note for JwtAuthCookieMiddleware usage.
2. Remove unused import missed in earlier clean-up.

ARCH-262

FYI: It is not important to get this deployed, but just wanted to complete this update.


robrap:
@douglashall: Since 2.0.1 was already tagged and released, I think it is a good practice to update the version.  I do not need to tag and release at this time.  I think it is fine if other changes come in later on this new version, or if the version gets updated to 2.1.0 or 3.0.0 in a later PR and is then tagged and released.

PR #55, opened 2018-11-15
ARCH-283: Enhance exception message for jwt_decode_handler.

robrap:
Temporarily, ecommerce is delegating to this jwt_decode_handler, with
its own fallback to allow for multiple issuers. Logging suppression is
required until ecommerce is down to a single issuer and can use this
jwt_decode_handler directly.  Ecommerce needs to suppress logging
because the exceptions for BadSignature are invalid due to the fallback
decoder that can in fact properly decode the token.

Note: this should be reverted as part of ARCH-276.

ARCH-283

FYI: The version was already updated to 2.0.2 in a past PR but not yet released.

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/decoder.py
+++ b/edx_rest_framework_extensions/auth/jwt/decoder.py
@@ -93,7 +93,7 @@ def _set_token_defaults(token):
         )
         if jwt_version.major > supported_version.major:
             logger.info('Token decode failed due to unsupported JWT version number [%s]', str(jwt_version))
-            raise jwt.InvalidTokenError
+            raise jwt.InvalidTokenError('JWT version number [%s] is unsupported', str(jwt_version))
 
     def _get_and_set_version(token):
         """


edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -170,7 +174,8 @@ def _decode_and_verify_token(token, jwt_issuer):
     token_issuer = decoded_token.get('iss')
     issuer_matched = any(issuer['ISSUER'] == token_issuer for issuer in get_jwt_issuers())
     if not issuer_matched:
-        logger.info('Token decode failed due to mismatched issuer [%s]', token_issuer)
+        if not suppress_logging:
+            logger.info('Token decode failed due to mismatched issuer [%s]', token_issuer)
         raise jwt.InvalidTokenError('%s is not a valid issuer.', token_issuer)
```
robrap: Reviewers: Codecov is reporting that we don't have coverage for an error with the invalid issuer.  I could add a test for that.  

Codecov seems to be happy with coverage of the rest, even though we aren't testing `suppress_logging = True`, but I'm happy with that.  I wasn't intending to add testing here anyway.  I did fix the test in ecommerce that wants this logging to be suppressed.

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -81,19 +83,20 @@ def decode_jwt_filters(token):
...
+    def _verify_version(jwt_version, suppress_logging=None):
         supported_version = Version(
             settings.JWT_AUTH.get('JWT_SUPPORTED_VERSION', JwtTokenVersion.latest_supported)
         )
         if jwt_version.major > supported_version.major:
-            logger.info('Token decode failed due to unsupported JWT version number [%s]', str(jwt_version))
-            raise jwt.InvalidTokenError
+            if not suppress_logging:
+                logger.info('Token decode failed due to unsupported JWT version number [%s]', str(jwt_version))
+            raise jwt.InvalidTokenError('JWT version number [%s] is unsupported', str(jwt_version))
```
robrap: Reviewers: FYI: I added the reason the token was invalid to the exception raised.

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -170,7 +168,6 @@ def _decode_and_verify_token(token, jwt_issuer):
     token_issuer = decoded_token.get('iss')
     issuer_matched = any(issuer['ISSUER'] == token_issuer for issuer in get_jwt_issuers())
     if not issuer_matched:
-        logger.info('Token decode failed due to mismatched issuer [%s]', token_issuer)
```
robrap: @nasthagiri: As noted in Slack, I could leave this info log if you think it is important, but I would imagine that the raised InvalidTokenError would get logged.  Same goes for the above info log message.  I'm good with leaving it or removing it.

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -139,7 +138,6 @@ def _verify_jwt_signature(token, jwt_issuer):
     try:
         _ = JWS().verify_compact(token, key_set)
     except Exception as exc:
-        logger.exception('Token verification failed.')
```
robrap: @nasthagiri: This is the only message that I really need/want to remove.  The others I removed for consistency, assuming that the callers would handle logging.

robrap:
@douglashall @nasthagiri: FYI: I'll need to update tests, but until we first agree on a plan. :)

nasthagiri:
If you haven’t confirmed that the .info messages get propagated and they aren’t a concern right now, then why not just leave them there?

robrap:
@nasthagiri: This is all set.

nasthagiri:
@robrap Since I was directly pinged on this, I'm sharing my opinion - however, I do feel we have already spent over budgeted time on this change - so I do leave it to you and @douglashall to close this out (since I'm at the conference again today) - however you two choose to do so as a team.

In my opinion, this is my preferred order of addressing this issue:
1. Reorder the decoder code in e-commerce.  We have more control of that code as we address concerns specific to e-commerce.  That code has a waffle switch that can conditionally enable more logging if/when needed.  I may be wrong, but I believe Django's Authentication layer does not propagate full error information up the stack (and that's why we added this log statement in the first place). So unless we confirm that I am wrong about this, deleting the log statement may make it harder for us to debug authn issues in the future.
2. Deleting the log statement as you have in this PR seems cleaner and less intrusive than the earlier version.  If we run into debugging authn issues in the future, we may need to update the code at that time (but only if Django doesn't properly propagate the error).  It's possible that the current log statement doesn't include sufficient information today anyway (since it doesn't capture the full stack trace).

edx_rest_framework_extensions/auth/jwt/decoder.py
robrap: No longer applies.

edx_rest_framework_extensions/auth/jwt/decoder.py
robrap: Left as-is.

edx_rest_framework_extensions/auth/jwt/decoder.py
robrap: Reverted this change.

robrap:
Reviewers: updated PR to be a minor enhancement I didn't want to lose.  I do not plan on releasing this at this time.

PR #63, opened 2019-02-05
ARCH-379: Remove version verification

robrap:
FYI: There are two ways to go here:
1. Leave this code and update the latest_supported version to 2.0.0, or
2. Drop the version checking.

I explain in the commit comment below why I chose #2.

IMPORTANT: I haven't completed the LMS side of this work.  At that point, it might make it more clear whether or not this was the right decision.  Reviewers should wait until that is ready.

-------------------------------------------

Separately, the LMS oauth_dispatch is being updated to be capable of
returning JWTs with different major versions when specifically
requested.

For the planned v1.x to v2.x JWT, the edx-drf-extensions _verify_version
call didn't actually help, and just forced an upgrade/deployment of
every service. Therefore, this check is being removed until/unless it
can be coded to solve a particular problem.

ARCH-379

### DIFF
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '2.0.3'  # pragma: no cover
+__version__ = '2.0.4'  # pragma: no cover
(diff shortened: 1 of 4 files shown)


robrap:
This still may be a worth while change, even with no other changes, but closing for now due to uncertainty.  This PR has been cataloged in a new ticket: [ARCH-398](https://openedx.atlassian.net/browse/ARCH-398).

PR #65, opened 2019-03-12
ARCH-418: Remove jwt_refresh_cookie

nasthagiri:
The login_refresh endpoint in the LMS was updated to be based on the
user's session, rather than requiring an additional refresh cookie, so
removing any references to the unused refresh cookie and
JWT_AUTH_REFRESH_COOKIE setting.

https://openedx.atlassian.net/browse/ARCH-418

### DIFF
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '2.4.3'  # pragma: no cover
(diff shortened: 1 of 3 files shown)


robrap:
@nasthagiri: Did you forget about this PR, or intentionally wanted to abandon this for some reason?

robrap:
The rest of the clean-up seems to be a little larger because of all the repos involved, so I created https://openedx.atlassian.net/browse/ARCH-1271

PR #66, opened 2019-03-14
Bumping latest_supported value of JwtTokenVersion to 1.2.0

christopappas:


edx_rest_framework_extensions/__init__.py
```diff
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '2.0.3'  # pragma: no cover
+__version__ = '2.0.4'  # pragma: no cover
```
robrap: I know it is a tiny change, but I usually reserve the last for bug fixes. That would make this 2.1.0.

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -15,7 +15,7 @@
 
 
 class JwtTokenVersion(object):
-    latest_supported = '1.1.0'
+    latest_supported = '1.2.0'
```
robrap: 1. I thought this is where we were going to comment a mini changelog of past versions, or is that somewhere in edx-platform?
2. Is this what we discussed becoming `default_latest_supported`?

edx_rest_framework_extensions/__init__.py
christopappas: can do

edx_rest_framework_extensions/auth/jwt/decoder.py
christopappas: I didn't know where we intended to put the comment. I can put it here

I wasn't sure we decided to change `default_latest_supported` based on the slack messages, so I didn't. I can though

christopappas: Actually i'm not going to add a mini changelog. i believe we determined git log was the best place for it

PR #69, opened 2019-05-07
ENT-1783: Update cookies.get_decoded_jwt logic to also query jwt cookie from request.auth

saleem-latif:
**Description:**
This PR updates the logic of `edx_rest_framework_extensions.auth.jwt.cookies.get_decoded_jwt` so that jwt cookie is queried in `request.auth` if not found in `request.COOKIES`.

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/cookies.py
+++ b/edx_rest_framework_extensions/auth/jwt/cookies.py
@@ -29,7 +29,7 @@ def get_decoded_jwt(request):
     Returns a decoded jwt dict if it can be found.
     Returns None if the jwt is not found.
     """
-    jwt_cookie = request.COOKIES.get(jwt_cookie_name(), None)
+    jwt_cookie = request.COOKIES.get(jwt_cookie_name(), None) or getattr(request, 'auth', None)
 
     if not jwt_cookie:
         return None
--- a/edx_rest_framework_extensions/auth/jwt/tests/test_cookies.py
+++ b/edx_rest_framework_extensions/auth/jwt/tests/test_cookies.py
@@ -46,6 +46,17 @@ class TestJwtAuthCookies(TestCase):
         self.assertEquals(expected_decoded_jwt, decoded_jwt)
 
     def test_get_decoded_jwt_when_no_cookie(self):
-        mock_request = mock.Mock(COOKIES={})
+        mock_request = mock.Mock(COOKIES={}, auth=None)
 
         self.assertIsNone(cookies.get_decoded_jwt(mock_request))
+
+    def test_get_decoded_jwt_from_auth(self):
+        user = UserFactory()
+        payload = generate_latest_version_payload(user)
+        jwt = generate_jwt_token(payload)
+        expected_decoded_jwt = jwt_decode_handler(jwt)
+
+        mock_request_with_cookie = mock.Mock(COOKIES={}, auth=jwt)
+
+        decoded_jwt = cookies.get_decoded_jwt(mock_request_with_cookie)
+        self.assertEquals(expected_decoded_jwt, decoded_jwt)
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,3 +2,4 @@ Clinton Blackburn <cblackburn@edx.org>
 Bill DeRusha <bill@edx.org>
 Jeremy Bowman <jbowman@edx.org>
 Robert Raposa <rraposa@edx.org>
+Saleem Latif <saleem_ee@hotmail.com>
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '2.2.0'  # pragma: no cover
+__version__ = '2.2.1'  # pragma: no cover


edx_rest_framework_extensions/auth/jwt/tests/test_cookies.py
```diff
@@ -46,6 +46,17 @@ def test_get_decoded_jwt_from_existing_cookie(self):
         self.assertEquals(expected_decoded_jwt, decoded_jwt)
 
     def test_get_decoded_jwt_when_no_cookie(self):
-        mock_request = mock.Mock(COOKIES={})
+        mock_request = mock.Mock(COOKIES={}, auth=None)
```
mushtaqak: I guess we won't need pass this since we are getting `getattr(request, 'auth', None` in `get_decoded_jwt` method.

saleem-latif: We need to mock it out otherwise a mock object is returned instead of a `None`. Since, `None or Mock()` would return `Mock()`.

edx_rest_framework_extensions/auth/jwt/cookies.py
```diff
@@ -29,7 +29,7 @@ def get_decoded_jwt(request):
     Returns a decoded jwt dict if it can be found.
     Returns None if the jwt is not found.
     """
-    jwt_cookie = request.COOKIES.get(jwt_cookie_name(), None)
+    jwt_cookie = request.COOKIES.get(jwt_cookie_name(), None) or getattr(request, 'auth', None)
```
nasthagiri: The current interface of this function is intended for retrieving JWTs just from the Cookie.  Both the docstring and the name of the module (cookies.py) imply this boundary/responsibility.  I'm afraid callers of this functionality will expect the JWT to be retrieved just from the cookie and this additional functionality will have unintended consequences.

Can we move this functionality to another (maybe new) method?  Also, why is this change needed?  There are separate flows for JWT retrieval.  Does it make sense to couple those multiple flows into a single interface?

FYI @brittneyexline @robrap

brittneyexline: you are right - the version of the function that existed in edx-rbac that we were using before was agnostic to how the jwt was getting fetched, and i didn't notice the subtle difference here wrt function naming/expectations. the reason we need to check `request.auth` for the jwt is for cases where users are going through the client credentials grant flow. for our rbac authz checks on some endpoints, this is the predominant way those endpoints were getting accessed, so we needed to get role data from the jwt which is only present in `request.auth` and not in the cookie.

i am ok with introducing a new function to get the jwt from `request.auth` and using this where we need it for our rbac authz checks. that way the current callers of this function can maintain their expectations of this function and we avoid this issue.

i still don't fully understand why `ecommerce_worker`'s jwt auth failed the token verification? i'd like to understand just for my own edification :-)

brittneyexline: i'm going to propose adding a similar function to get the decoded jwt from `request.auth` to https://github.com/edx/edx-drf-extensions/blob/master/edx_rest_framework_extensions/auth/jwt/authentication.py, do you think that is an appropriate place for this function?

robrap: @brittneyexline: That seems like the right place.  Some pseudo-code for the new function might be:
```
def get_decoded_jwt(request):
    return is_jwt_authenticated(request) ? getattr(request, 'auth', None) | None
```
Also, you may want to look at Splunk to confirm that we aren't seeing [this error log](https://github.com/edx/edx-drf-extensions/blob/master/edx_rest_framework_extensions/auth/jwt/authentication.py#L86).  That was added to ensure that this would be available in all cases, but not sure if it is.

robrap: @brittneyexline: To answer your question about what went wrong, it starts in this call to [`_get_lms_user_id_from_jwt_cookie`](https://github.com/edx/ecommerce/blob/cc496b0524b34a8667f2a7b105101e3c41b2a922/ecommerce/core/models.py#L514), where it tried to decode the JWT cookie.  This code should probably be updated to use your new method (when you add it) and just have access to the decoded JWT off the bat, and to all JWTs, no matter where it comes from, as long as it passes authentication.
FYI: @crice100: You may want to consider making this small update as part of your work since you are in there.  Let me know if you need more explanation.

robrap: Also, @brittneyexline, if you want to clean-up the versioning, I would revert the change that was made and release 2.2.2 (to clean-up the 2.2.x versions), and then release your new change as 2.3.0.

nasthagiri: Thanks @robrap.

Regarding the Splunk confirmation, I had done so earlier and found that that error statement doesn't exist in our logs.  We had added that log earlier to confirm our understanding of how Django's authentication stores JWT in the request.  Given that it hasn't been logged, we can more safely assume that `request.auth` exists when `is_jwt_authenticated` is `True`.

brittneyexline: thanks @robrap for the info/suggestions!

@saleem-latif - in addition to the info on the ticket, please note the suggested fix for cleaning up versioning - if you could do that before proceeding with the other changes, that would be much appreciated!

PR #72, opened 2019-05-24
ENT-1783: Added a helper method to get jwt token from request.auth

saleem-latif:
**Description:**
This PR adds a helper method in `edx_rest_framework_extensions/auth/jwt/authentication.py` to get decoded hwt token from `request.auth`


edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -6,6 +6,8 @@
 from rest_framework import exceptions
 from rest_framework_jwt.authentication import JSONWebTokenAuthentication, BaseJSONWebTokenAuthentication
 
+from edx_rest_framework_extensions.auth.jwt.cookies import jwt_cookie_name
```
brittneyexline: this looks like an unused import

PR #76, opened 2019-07-01
Add MiddlewareMixin to all middleware classes

kdmccormick:
Django >=1.10 uses a slightly different middleware framework. Using the old framework results in RemovedInDjango20 warnings. Django provides MiddlewareMixin for classes that work as both old-style and new-style middlewares. Inheriting from it allows clients of this library to move to the new middleware framwork and silence the deprecation warnings.


kdmccormick:
Thanks @mikix. Looking at those docs, it seems like the only difference is exception handling, which shouldn't affect these classes.

PR #81, opened 2019-07-19
Allow specifying certain JWT claims to merge into user attributes, ra…

cpennington:
…ther than overwriting them


cpennington:
I was actually just in the process of adding a test.

PR #86, opened 2019-08-09
ARCH-1051: enables this to work with jwt auth login required middleware

robrap:
Middleware enables DRF JwtAuthentication authentication class for
endpoints using the new LoginRequired permission class.

Enables a non-standard use of DRF, because DRF only supports returning
a 401 for unauthorized users, but this middleware allows a DRF endpoint
to redirect to login.

This can be used to convert a plain Django view which was using
@login_required into a DRF APIView, which is useful to enable our DRF
JwtAuthentication authentication class.

NOTE: This includes a breaking change that is unlikely to affect anyone
unless they subclassed `JwtAuthCookieMiddleware`, which switched from
using `process_request` to `process_view` so it would not run before
this new middleware.

ARCH-1051

The ecommerce PR that makes use of this middleware can be found here:
https://github.com/edx/ecommerce/pull/2533

Note: `JwtAuthCookieMiddleware` already includes a metric for whether or not the cookies are available, so we shouldn't need an additional metric.

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/middleware.py
+++ b/edx_rest_framework_extensions/auth/jwt/middleware.py
@@ -3,8 +3,10 @@ Middleware supporting JWT Authentication.
 """
 import logging
 
+from django.contrib.auth.decorators import login_required
 from django.utils.deprecation import MiddlewareMixin
 from edx_django_utils import monitoring
+from edx_django_utils.cache import RequestCache
 from rest_framework_jwt.authentication import BaseJSONWebTokenAuthentication
 
 from edx_rest_framework_extensions.auth.jwt.cookies import (
@@ -13,7 +15,7 @@ from edx_rest_framework_extensions.auth.jwt.cookies import (
     jwt_cookie_signature_name,
 )
 from edx_rest_framework_extensions.auth.jwt.constants import JWT_DELIMITER
-from edx_rest_framework_extensions.permissions import NotJwtRestrictedApplication
+from edx_rest_framework_extensions.permissions import LoginRedirectIfUnauthenticated, NotJwtRestrictedApplication
 
 log = logging.getLogger(__name__)
 USE_JWT_COOKIE_HEADER = 'HTTP_USE_JWT_COOKIE'
@@ -25,14 +27,6 @@ class EnsureJWTAuthSettingsMiddleware(MiddlewareMixin):
     are set on all endpoints that use JWTAuthentication.
     """
     _required_permission_classes = (NotJwtRestrictedApplication,)
-
-    def _includes_base_class(self, iter_classes, base_class):
-        """
-        Returns whether any class in iter_class is a subclass of the given base_class.
-        """
-        return any(
-            issubclass(auth_class, base_class) for auth_class in iter_classes,
-        )
 
     def _add_missing_jwt_permission_classes(self, view_class):
         """
@@ -56,7 +50,7 @@ class EnsureJWTAuthSettingsMiddleware(MiddlewareMixin):
                     view_permissions.append(child)
 
         for perm_class in self._required_permission_classes:
-            if not self._includes_base_class(permission_classes, perm_class):
+            if not _includes_base_class(permission_classes, perm_class):
                 message = (
                     u"The view %s allows Jwt Authentication. The required permission class, %s,",
                     u" was automatically added."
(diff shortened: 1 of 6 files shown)


edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -122,7 +122,7 @@ def _get_missing_cookie_message_and_metric(self, cookie_name):
         request_jwt_cookie = 'missing-{}'.format(cookie_name)
         return cookie_missing_message, request_jwt_cookie
 
-    def process_request(self, request):
+    def process_view(self, request, view_func, view_args, view_kwargs):  # pylint: disable=unused-argument
```
robrap: @douglashall: Do you see any issue with this particular change?  The `process_request` call was happening for all middleware before `process_view`, so when I tried introducing another middleware that could change the request before this one, but needed to use `process_view`, it was getting called too late.  Updating this to `process_view` allows the other one to set the header.  

See https://github.com/edx/ecommerce/pull/2531/files . I do plan on moving this middleware into this file (and probably PR), so it will eventually be more understandable.

robrap: @douglashall: Please ignore this comment for now.  This PR will be updated.

robrap: I'm marking my own question as resolved.

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -13,27 +15,39 @@
...
 )
 from edx_rest_framework_extensions.auth.jwt.constants import JWT_DELIMITER
-from edx_rest_framework_extensions.permissions import NotJwtRestrictedApplication
+from edx_rest_framework_extensions.permissions import LoginRequired, NotJwtRestrictedApplication
 
 log = logging.getLogger(__name__)
 USE_JWT_COOKIE_HEADER = 'HTTP_USE_JWT_COOKIE'
 
 
+def _includes_base_class(iter_classes, base_class):
```
nasthagiri: Nit: I've typically been following the top-down rule from the `Clean Code` book for readability.  So that the helper functions/methods are below the higher-level public methods.

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -72,19 +86,87 @@ def _add_missing_jwt_permission_classes(self, view_class):
...
+        if _includes_base_class(view_authentication_classes, BaseJSONWebTokenAuthentication):
             self._add_missing_jwt_permission_classes(view_class)
 
 
+class JwtAuthWithLoginRequiredMiddleware(MiddlewareMixin):
+    """
+    Middleware enables DRF JwtAuthentication authentication class for endpoints
+    using the LoginRequired permission class.
+
+    Enables a non-standard use of DRF, because DRF only supports returning
```
nasthagiri: Does it make sense to classify this as 'non-standard'?

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -72,19 +86,87 @@ def _add_missing_jwt_permission_classes(self, view_class):
...
+    Usage Notes:
+    - This middleware must be added before JwtAuthCookieMiddleware.
+    - Only affects endpoints using the LoginRequired permission class.
+
+    See https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/oauth_dispatch/docs/decisions/0009-jwt-in-session-cookie.rst  # noqa
+    """
+    _LOGIN_REQUIRED_FOUND_CACHE_KEY = 'login_required_found'
+
+    def _get_request_cache(cls):
+        cache_namespace = 'JwtAuthWithLoginRequiredMiddleware'
```
nasthagiri: Nit: define this as a constant, along with `_LOGIN_REQUIRED_FOUND_CACHE_KEY`.

edx_rest_framework_extensions/permissions.py
```diff
@@ -144,6 +144,26 @@ def _get_user_filter(self, request):
...
 
 
+class LoginRequired(IsAuthenticated):
+    """
+    A non-standard DRF permission class that will login redirect unauthorized users.
+
+    It can be used to convert a plan Django view using @login_required into a
+    DRF APIView, which is useful to enable our DRF JwtAuthentication class.
+
+    This permission is non-standard because DRF only supports returning a 401
```
nasthagiri: Once again, not sure how `non-standard` this is.  The `@login_required` decorator redirected to login as well.

edx_rest_framework_extensions/auth/jwt/tests/test_urls.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import absolute_import
```
nasthagiri: Can this be colocated and defined within the test file itself?    If not, can we rename this file so it's clear that it doesn't include tests itself?  The test runner will also expect this named file to contain tests.

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -72,19 +86,87 @@ def _add_missing_jwt_permission_classes(self, view_class):
...
+        if self.is_jwt_auth_enabled_with_login_required(request, view_func):
+            request.META[USE_JWT_COOKIE_HEADER] = 'true'
+
+    def process_response(self, request, response):
+        """
+        Redirects unauthenticated users to login when LoginRequired permission class was used.
+        """
+        if self._is_login_required_found() and not request.user.is_authenticated:
+            login_url = self.get_login_url(request)
+            return login_required(function=lambda request: None, login_url=login_url)(request)
```
nasthagiri: Nice find in reusing the `login_required` Django function.  BTW, the existence of this function illustrates that Django supports redirecting to login.

nasthagiri:
@robrap This looks great!  I like how you cleaned it up and made it generic as well!  Awesome.

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: I'll see if it is simple enough.  I think I prefer what I did out of habit, not because it is better. :)

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: It's non-standard when it comes to DRF.  That is why we need special middleware.

Do you have a proposal that is better and strong enough?  I think this is the right solution in certain situations, but not many.

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: It originally had that, but it made the code messier for a constant that wasn't shared.  I wish it was fine in Python to use the following in a method:
```
CACHE_NAMESPACE = 'JwtAuthWithLoginRequiredMiddleware'
```
Of the two imperfect options: 1) all caps at the class level, or 2) lower case variable, I still prefer what I have.

edx_rest_framework_extensions/permissions.py
robrap: See other comment.

edx_rest_framework_extensions/auth/jwt/tests/test_urls.py
robrap: I would like to co-locate if possible.  I was just trying to get it to work at first.  I'll try that.

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: Django definitely does.  DRF does not. :)

edx_rest_framework_extensions/auth/jwt/middleware.py
nasthagiri: I suggested it, mostly for readability, so that the cache-namespace string is co-located with the cache-key string.

edx_rest_framework_extensions/auth/jwt/tests/test_urls.py
nasthagiri: You could also rename this file to `url_patterns.py` and keep it in the `tests` folder.

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -72,19 +86,87 @@ def _add_missing_jwt_permission_classes(self, view_class):
...
+class JwtAuthWithLoginRequiredMiddleware(MiddlewareMixin):
+    """
+    Middleware enables DRF JwtAuthentication authentication class for endpoints
+    using the LoginRequired permission class.
+
+    Enables a non-standard use of DRF, because DRF only supports returning
+    a 401 for unauthorized users, but this middleware allows a DRF endpoint
+    to redirect to login.
+
+    This can be used to convert a plan Django view using @login_required into a
```
nasthagiri: `plan` -> `plain`

edx_rest_framework_extensions/permissions.py
```diff
@@ -144,6 +144,26 @@ def _get_user_filter(self, request):
         return None
 
 
+class LoginRequired(IsAuthenticated):
```
nasthagiri: Consider renaming this to `LoginRedirectIfUnauthenticated` class.

nasthagiri: Nit: It's kind of odd that this `Permission` class is very far away from its sister `Authentication` class.

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -72,19 +86,87 @@ def _add_missing_jwt_permission_classes(self, view_class):
...
-        )
+        view_class = _get_view_class(view_func)
 
         view_authentication_classes = getattr(view_class, 'authentication_classes', tuple())
-        if self._includes_base_class(view_authentication_classes, BaseJSONWebTokenAuthentication):
+        if _includes_base_class(view_authentication_classes, BaseJSONWebTokenAuthentication):
             self._add_missing_jwt_permission_classes(view_class)
 
 
+class JwtAuthWithLoginRequiredMiddleware(MiddlewareMixin):
```
nasthagiri: Consider renaming this to: `JwtRedirectToLoginIfUnauthenticatedMiddleware`.  Such a name may make it explicit on what it enables.

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -72,19 +86,87 @@ def _add_missing_jwt_permission_classes(self, view_class):
...
+    """
+    Middleware enables DRF JwtAuthentication authentication class for endpoints
+    using the LoginRequired permission class.
+
+    Enables a non-standard use of DRF, because DRF only supports returning
+    a 401 for unauthorized users, but this middleware allows a DRF endpoint
```
nasthagiri: Note: `401` should be returned for "unauthenticated" users, and `403` should be returned for "unauthorized" users.

edx_rest_framework_extensions/auth/jwt/middleware.py
nasthagiri: Consider rephrasing this to:
```
Enables a DRF view to redirect the user to login when they are unauthenticated.  It automatically enables JWT-cookie-based authentication by setting the `USE_JWT_COOKIE_HEADER`.
```

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: Done.

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: See separate rephrasing comment.

robrap: Added note to end about LoginRequired.

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: Yup.  Thanks.  Deleted anyway based on your rephrasing.

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: Done.

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: Done.  I moved it lower, so less of an issue.

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: We've rephrased things, so I think we are all set.

edx_rest_framework_extensions/auth/jwt/tests/test_urls.py
robrap: Done.

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: Done.

edx_rest_framework_extensions/permissions.py
robrap: Done.

robrap: 100% Agree.  However, I thought the best short-term was to put it with all the other JWT based permission classes.

Note: We did a big refactor of edx-drf-extensions, but the refactor got too large, so we decided to put off refactoring the permissions.  I think we should finish that work at some point, but not as part of this PR.

edx_rest_framework_extensions/permissions.py
robrap: Fixed.

PR #87, opened 2019-08-14
chore: use AGPLv3 instead of GPLv3

albemarle:


nedbat:
According to our licensing guidelines, I think this repo should be Apache: https://openedx.atlassian.net/wiki/spaces/OPEN/pages/13959332/Licensing

PR #88, opened 2019-08-15
chore: use Apache 2 license instead of GPLv3

albemarle:


nedbat:
Great! Also update setup.py: https://github.com/edx/edx-drf-extensions/blob/master/setup.py#L15-L20

nedbat:
And the readme: https://github.com/edx/edx-drf-extensions/blame/master/README.rst#L33

albemarle:
@nedbat Updated those places (though it seems to me we could take it out of the README...).

setup.py
```diff
@@ -12,7 +12,7 @@
     author='edX',
     author_email='oscm@edx.org',
     url='https://github.com/edx/edx-drf-extensions',
-    license='AGPL',
+    license='Apache 2.0',
     classifiers=[
         'Development Status :: 5 - Production/Stable',
         'Environment :: Web Environment',
```
nedbat: Sorry to be persnickety, but two lines down, it should say:
```
    'License :: OSI Approved :: Apache Software License',
```

albemarle: All set!

PR #89, opened 2019-10-04
ARCH-1197: request.user set in JwtAuthCookieMiddleware

robrap:
JWT cookie authentication uses DRF, which does not set `request.user`
until the Middleware call: `process_response`.  This change sets the
`request.user` in process_view, which happens after all
`process_request` calls.

This is has been added to the existing `JwtAuthCookieMiddleware`.

ARCH-1197

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/tests/test_middleware.py
+++ b/edx_rest_framework_extensions/auth/jwt/tests/test_middleware.py
@@ -5,9 +5,10 @@ import ddt
 from django.conf.urls import url
 from django.http.cookie import SimpleCookie
 from django.test import Client, RequestFactory, TestCase, override_settings
+from django.utils.deprecation import MiddlewareMixin
 from edx_django_utils.cache import RequestCache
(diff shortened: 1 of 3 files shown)


edx_rest_framework_extensions/auth/jwt/tests/test_middleware.py
```diff
@@ -358,3 +372,27 @@ def test_success(self, mock_set_custom_metric):
...
+        """
+        header = {USE_JWT_COOKIE_HEADER: 'true'}
+        self.client.cookies = SimpleCookie({
+            jwt_cookie_header_payload_name(): 'header.payload',
+            jwt_cookie_signature_name(): 'signature',
+        })
+        response = self.client.get('/nopermissionsrequired/', **header)
+        self.assertEqual(200, response.status_code)
+
+    def _mock_jwt_auth_authenticate(self, request):
```
dianakhuang: Do we need this for anything? I don't see it referenced anywhere.

PR #90, opened 2019-10-15
ARCH-1210: Add toggle to determine whether to set request.user.

robrap:
Although setting the `request.user` during `process_view` worked to
solve ARCH-1197 in ecommerce, it introduced an OOM bug in edx-platform.

This introduces a temporary toggle using a Django Setting so that the
fix can remain in ecommerce without affecting other IDAs like
edx-platform.

Once ARCH-1199 is complete and the issue is fixed, it is expected that
this toggle will be removed.

ARCH-1210

**UPDATE:** The second commit refactors from a Waffle Switch to a Django Setting so that we are not blocked by Devops, since we don't have access to Ecommerce Admin in Production.  It also better allows me to check that the solution sticks for ecommerce.

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/middleware.py
+++ b/edx_rest_framework_extensions/auth/jwt/middleware.py
@@ -12,6 +12,7 @@ from edx_django_utils.cache import RequestCache
 from rest_framework.request import Request
 from rest_framework_jwt.authentication import BaseJSONWebTokenAuthentication
 
+from edx_rest_framework_extensions.config import ENABLE_SET_REQUEST_USER_FOR_JWT_COOKIE
 from edx_rest_framework_extensions.auth.jwt.cookies import (
     jwt_cookie_name,
     jwt_cookie_header_payload_name,
@@ -19,6 +20,7 @@ from edx_rest_framework_extensions.auth.jwt.cookies import (
 )
 from edx_rest_framework_extensions.auth.jwt.constants import JWT_DELIMITER
 from edx_rest_framework_extensions.permissions import LoginRedirectIfUnauthenticated, NotJwtRestrictedApplication
+from edx_rest_framework_extensions.settings import get_setting
 
 log = logging.getLogger(__name__)
 USE_JWT_COOKIE_HEADER = 'HTTP_USE_JWT_COOKIE'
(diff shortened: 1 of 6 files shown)


robrap:
@nasthagiri: Adding the waffle switch did force me to clean up the tests a little. :)

edx_rest_framework_extensions/auth/jwt/tests/test_middleware.py
```diff
@@ -2,6 +2,7 @@
 Unit tests for jwt authentication middlewares.
 """
 import ddt
+import waffle
```
nasthagiri: unused import?

edx_rest_framework_extensions/config.py
```diff
@@ -2,6 +2,18 @@
 Application configuration constants and code.
 """
 
-NAMESPACE_SWITCH = 'oauth2'
-SWITCH_ENFORCE_JWT_SCOPES = 'enforce_jwt_scopes'
-NAMESPACED_SWITCH_ENFORCE_JWT_SCOPES = '{}.{}'.format(NAMESPACE_SWITCH, SWITCH_ENFORCE_JWT_SCOPES)
+OAUTH_TOGGLE_NAMESPACE = 'oauth2'
+SWITCH_ENFORCE_JWT_SCOPES = '{}.enforce_jwt_scopes'.format(OAUTH_TOGGLE_NAMESPACE)
+
+# .. toggle_name: EDX_DRF_EXTENSIONS: ENABLE_SET_REQUEST_USER_FOR_JWT_COOKIE
```
nasthagiri: Curious: Are toggle names intended to be prefixed with the repo name?

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -3,6 +3,7 @@
 """
 import logging
 
+import waffle
```
nasthagiri: another unused import?

edx_rest_framework_extensions/auth/jwt/tests/test_middleware.py
robrap: Fixed.  Thanks.

edx_rest_framework_extensions/config.py
robrap: 1. Adding a new Django Setting following an existing pattern which namespaces in this way.
2. In general, I like this pattern to help people find settings more easily.  Note that I don't know if our toggle report is yet looking at dependency repos.

PR #91, opened 2019-11-01
Add API docs tooling

kdmccormick:
@nedbat @regisb -- Here's my stab at pulling the API doc tools out of `edx-platform` and into a shared repository. I settled on `edx-drf-extensions` because it's our canonical "REST API tooling" library and is already installed into most edX IDAs.

This PR creates a new folder called `edx_api_doc_tools` that exposes a toolkit of utilities through its `__init__.py`. It also has a silly "edX Hedgehog API" under `edx_api_doc_tools/example` that demonstrates how to use the doc tools. Finally, `edx_api_doc_tools/tests` does some baseline testing by looking at the JSON generated by from the example API.

There are a few `TODO`s scattered throughout this PR (hence its DRAFT status), but I wanted to solicit feedback early in case there are any major changes I need to make.

I haven't included Sphinx doc generation in this PR; I'm thinking that it could merge separately, while applying this PR's changes to `edx-platform` could be done in parallel.

Lastly, @regis and @nedbat , I plan on giving you both co-authorship on this final commit(s), as you wrote much of the original code; let me know if you have any qualms with that :)

### DIFF
--- a/edx_api_doc_tools/conf_utils.py
+++ b/edx_api_doc_tools/conf_utils.py
@@ -1,0 +1,148 @@
+"""
+Functions for setting up API doc generation & viewing in a repository.
+
+External users: import these from __init__.
+"""
+from __future__ import absolute_import, unicode_literals
+
+from django.conf import settings
+from django.conf.urls import url
+from django.views.generic.base import RedirectView
+from drf_yasg import openapi
+from drf_yasg.generators import OpenAPISchemaGenerator
+from drf_yasg.views import get_schema_view
+from rest_framework import permissions
+
+
+def make_docs_urls(api_info):
+    """
+    Convenience function for creating API doc views given an API info object.
+
+    Arguments:
+        api_info (openapi.Info)
+
+    Returns: list[RegexURLPattern]
+        A list of url patterns to the API docs.
+    """
+    return get_docs_urls(
+        docs_data_view=make_docs_data_view(api_info),
+        docs_ui_view=make_docs_ui_view(api_info),
+    )
+
+
+def make_api_info(
+        title="Open edX APIs",
+        version="v1",
+        email="oscm@edx.org",
+        description="APIs for access to Open edX information",
+):
+    """
+    Build an API info object.
+
+    Arguments:
+        * title (str)
+        * version (str)
+        * email (str)
+        * description (str)
+
+    Returns: openapi.Info
+    """
+    return openapi.Info(
+        title=title,
+        default_version=version,
+        contact=openapi.Contact(email=email),
+        description=description,
+    )
+
+
+def get_docs_urls(docs_data_view, docs_ui_view):
+    """
+    Get some reasonable URL patterns to browsable API docs and API docs data.
+
+    If these URL patterns don't work for your service,
+    feel free to construct your own.
+
+    Arguments:
+        docs_data_view (openapi.Info): JSON/YAML view for API docs data.
+        docs_ui_view (openapi.Info): Nice HTML view for API docs.
+
+    Returns: list[RegexURLPattern]
+        A list of url patterns to the API docs.
+    """
+    return [
+        url(
+            r'^swagger(?P<format>\.json|\.yaml)$',
+            docs_data_view,
+            name='apidocs-data',
+        ),
+        url(
+            r'^api-docs/$',
+            docs_ui_view,
+            name='apidocs-ui',
+        ),
+        url(
+            r'^swagger/$',
+            RedirectView.as_view(pattern_name='apidocs-ui', permanent=False),
+            name='apidocs-ui-swagger',
+        ),
+    ]
+
+
+def make_docs_data_view(api_info):
+    """
+    Build View for API documentation data (either JSON or YAML).
+
+    Arguments:
+        api_info (openapi.Info)
+
+    Returns: View
+    """
+    return get_schema_view(
+        api_info,
+        generator_class=ApiSchemaGenerator,
+        public=True,
+        permission_classes=(permissions.AllowAny,),
+    ).without_ui(cache_timeout=get_docs_cache_timeout())
+
+
+def make_docs_ui_view(api_info):
+    """
+    Build View for browsable API documentation.
+
+    Arguments:
+        api_info (openapi.Info)
+
+    Returns: View
+    """
+    return get_schema_view(
+        api_info,
+        generator_class=ApiSchemaGenerator,
+        public=True,
+        permission_classes=(permissions.AllowAny,),
+    ).with_ui('swagger', cache_timeout=get_docs_cache_timeout())
+
+
+class ApiSchemaGenerator(OpenAPISchemaGenerator):
+    """
+    A schema generator for /api/*
+
+    Only includes endpoints in the /api/* url tree, and sets the path prefix
+    appropriately.
+    """
+    def get_endpoints(self, request):
+        endpoints = super(ApiSchemaGenerator, self).get_endpoints(request)
+        subpoints = {p: v for p, v in endpoints.items() if p.startswith("/api/")}
+        return subpoints
+
+    def determine_path_prefix(self, paths):
+        return "/api/"
+
+
+def get_docs_cache_timeout():
+    """
+    Return OPENAPI_CACHE_TIMEOUT setting, or zero if it's not defined.
+    """
+    try:
+        return settings.OPENAPI_CACHE_TIMEOUT
+    except AttributeError:
+        return 0
--- a/edx_api_doc_tools/conf_utils.py
+++ b/edx_api_doc_tools/conf_utils.py
@@ -23,6 +23,13 @@ def make_docs_urls(api_info):
 
     Returns: list[RegexURLPattern]
         A list of url patterns to the API docs.
+
+    Example:
+        # File: urls.py
+        from edx_api_doc_tools import make_docs_urls, make_api_info
+        urlpatterns = [ ... ] # Your URL patterns.
+        api_info = make_api_info(title="Awesome API", version="v42")
+        urlpatterns += make_docs_urls(api_info)
     """
     return get_docs_urls(
         docs_data_view=make_docs_data_view(api_info),
@@ -68,6 +75,13 @@ def get_docs_urls(docs_data_view, docs_ui_view):
 
     Returns: list[RegexURLPattern]
         A list of url patterns to the API docs.
+
+    Example:
+        # File: urls.py
+        from edx_api_doc_tools import get_docs_urls
+        from .views import custom_doc_data_view, custom_doc_ui_view
+        urlpatterns = [ ... ] # Your URL patterns.
+        urlpatterns += get_docs_urls(custom_doc_data_view, custom_doc_ui_view)
     """
     return [
         url(
@@ -96,6 +110,11 @@ def make_docs_data_view(api_info):
         api_info (openapi.Info)
 
     Returns: View
+
+    Example:
+        from edx_api_doc_tools import make_api_info, make_docs_data_view
+        api_info = make_api_info(title="Awesome API", version="v42")
+        my_data_view = make_docs_data_view(api_info)
     """
     return get_schema_view(
         api_info,
@@ -113,6 +132,11 @@ def make_docs_ui_view(api_info):
         api_info (openapi.Info)
 
     Returns: View
+
+    Example:
+        from edx_api_doc_tools import make_api_info, make_docs_ui_view
+        api_info = make_api_info(title="Awesome API", version="v42")
+        my_ui_view = make_docs_ui_view(api_info)
     """
     return get_schema_view(
         api_info,
--- a/edx_api_doc_tools/data.py
+++ b/edx_api_doc_tools/data.py
@@ -1,0 +1,95 @@
+"""
+Data definitions for API schemas.
+
+Mostly, we just use the definitions in drf_yasg.openapi
+"""
+from __future__ import absolute_import, unicode_literals
+
+from drf_yasg import openapi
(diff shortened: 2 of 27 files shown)


edx_api_doc_tools/conf_utils.py
```diff
@@ -0,0 +1,148 @@
...
+from django.views.generic.base import RedirectView
+from drf_yasg import openapi
+from drf_yasg.generators import OpenAPISchemaGenerator
+from drf_yasg.views import get_schema_view
+from rest_framework import permissions
+
+
+def make_docs_urls(api_info):
+    """
+    Convenience function for creating API doc views given an API info object.
```
nedbat: This docstring could use an example of how this function is used, how is it helpful? Ditto for the other functions in this module.

edx_api_doc_tools/conf_utils.py
```diff
@@ -0,0 +1,148 @@
...
+        * version (str)
+        * email (str)
+        * description (str)
+
+    Returns: openapi.Info
+    """
+    return openapi.Info(
+        title=title,
+        default_version=version,
+        contact=openapi.Contact(email=email or "ocsm@edx.org"),
```
nedbat: No need to default the email twice.

edx_api_doc_tools/data.py
```diff
@@ -0,0 +1,100 @@
+"""
+Data definitions for API schemas.
+
+These are generally thin wrappers around drf-yasg's
+OpenAPI schema types. We wrap them so that we can
+easily switch from drf-yasg to another supporting
+library if we must without affecting clients of this
+library.
```
nedbat: Why do you wrap docstrings and comments so narrowly?

edx_api_doc_tools/example/test_example.py
```diff
@@ -0,0 +1,86 @@
...
+        assert len(response.data) == 5
+
+    def test_get(self):
+        response = self.request('get', 'hogs/pepper/')
+        assert response.status_code == 200
+        assert response.data['name'] == 'Pepper Pokes'
+
+    def test_post(self):
+        response = self.request('post', 'hogs/', self.new_hog)
+        assert response.status_code == 501
```
nedbat: Why is this 501?

edx_api_doc_tools/example/views.py
```diff
@@ -0,0 +1,102 @@
...
+        query_parameter('min-grams', int, "Filter on whether minimum weight (grams)."),
+        query_parameter('min-ounces', float, "Filter hogs by minimum weight (ounces)."),
+        query_parameter('fav-food', str, "Filter hogs by favorite food."),
+        query_parameter('graduated', bool, "Filter hogs by whether they graudated."),
+    ],
+)
+@schema_for(
+    'retrieve',
+    # TODO: The below string becomes the operation's description instead
+    #       of its summary for some reason.
```
nedbat: Something wrong with the multi-line decision making?

edx_api_doc_tools/example/views.py
```diff
@@ -0,0 +1,102 @@
...
+                hog for hog in self.get_queryset()
+                if hog.key == hedgehog_key
+            )
+        except StopIteration:
+            raise NotFound()
+        return hedgehog
+
+    def perform_create(self, serializer):
+        """
+        TODO
```
nedbat: Do we have to implement these? Just for a test API, do we need those methods? Seems like that would be testing DRF, not our code.

edx_api_doc_tools/example/test_example.py
nedbat: Oh: because the method raises NotImplementedSomethingOrOther.

edx_api_doc_tools/tests/test_doc_tools.py
```diff
@@ -0,0 +1,64 @@
...
+
+    If you make a change to the doc tools or the example API that change the
+    generated schema, here's how to update to update `test_get_data_view`:
+    * Run the test, which should fail with an AssertionError about the
+      generated schema.
+    * Copy the contents of actual_schema.json into https://editor.swagger.io
+    * Make sure the browsable API looks correct,
+      and make sure the contents of the schema file look sane.
+    * If everything looks right, copy actual_schema.json into expected_schema.json
+      and commit the change, making it the new reference schema.
```
nedbat: I like these instructions :)

edx_api_doc_tools/tests/test_doc_tools.py
```diff
@@ -0,0 +1,64 @@
...
+            json.dump(actual_schema, f, indent=4)
+        assert actual_schema == self.expected_schema, (
+            "Generated schema (dumped to actual_schema.json) "
+            "did not match schema loaded from expected_schema.json."
+        )
+
+    def test_get_ui_view(self):
+        response = self.client.get('/api-docs/')
+        assert response.status_code == 200
+        assert 'edX Hedgehog Service API' in response.content.decode('utf-8')
```
nedbat: You mention wanting to beef this up: I think asserting about a handful of other strings will be good.

edx_api_doc_tools/tests/test_doc_tools.py
```diff
@@ -0,0 +1,64 @@
...
+      and make sure the contents of the schema file look sane.
+    * If everything looks right, copy actual_schema.json into expected_schema.json
+      and commit the change, making it the new reference schema.
+    """
+    base_path = dirname(__file__)
+    path_of_expected_schema = join(base_path, 'expected_schema.json')
+    path_of_actual_schema = join(base_path, 'actual_schema.json')
+
+    @classmethod
+    def setUpClass(cls):
```
nedbat: setUpClass seems overkill here, since you only use the expected schema in one test. Maybe you had thought there would be more?

edx_api_doc_tools/view_utils.py
```diff
@@ -0,0 +1,125 @@
...
+    class body, but is instead defined somewhere up in the DRF view hierarchy.
+    (For applying a schema directly to a method, use `schema`).
+
+    Valid values of `method_name` include:
+        'list',
+        'get', 'GET', 'retrieve',
+        'post', 'POST', 'create',
+        'put', 'PUT',
+        'patch', 'PATCH', 'update',
+        'delete', 'DELETE', and 'destroy'.
```
nedbat: Why does this decorator care what the method name is?  Maybe this should say, "DRF method names include:"

edx_api_doc_tools/view_utils.py
```diff
@@ -0,0 +1,125 @@
...
+    Arguments:
+        responses: dict[int, object]
+
+    Returns: dict[int, object]
+    """
+    if six.PY3 or responses is None:
+        return responses
+    return {
+        http_status: (
+            str(value)
```
nedbat: This will fail for non-ascii data, right? Shouldn't it be `value.encode("utf-8")` ?

edx_rest_framework_extensions/__init__.py
```diff
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '2.4.6'  # pragma: no cover
+__version__ = '3.0.0'  # pragma: no cover
```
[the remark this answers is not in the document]
kdmccormick: There was none, but I propose bumping it anyway, for no other reason than it'd be easy to remember that "Version 3 has the doc tools"

edx_api_doc_tools/example/views.py
kdmccormick: Yeah, I'm debugging it currently.

edx_api_doc_tools/tests/test_doc_tools.py
kdmccormick: Yep, that was it. I'm going to break this into function tests.

edx_api_doc_tools/view_utils.py
kdmccormick: It doesn't, but `swagger_auto_schema` only works on those names. Regardless, I like your rewording.

edx_rest_framework_extensions/__init__.py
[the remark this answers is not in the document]
kdmccormick: "Version 2.5 has the doc tools" is already fairly memorable, I guess.

edx_api_doc_tools/data.py
kdmccormick: Bad habit, or an overreaction to seeing 120+-wide unwrapped docstrings, I guess. I'll try hard to wrap at column ~80.

regisb:
Hey @kdmccormick! Good job working on this. The tests did not exist before in edx-platform, right? I can see that a lot of effort has gone into this.
I have qualms about the integration with the edx-drf-extensions repo: 

1. The large number of changes made to the requirements files show that the edx_api_doc_tools don't have much in common with the rest of this repo.
2. Unless I'm mistaken, edx_api_doc_tools does not depend on the rest of the code in the repo. Coalescing both repos would introduce an unnecessary dependency.
3. This PR adds 1k+ lines of code to a repo that has just 2.3k lines of code (roughly). This seems disproportionate.

Also, we should try to make it as easy as possible to use the API documentation tools. For that, we need to provide some documentation on how to use it. It would be great if we could have a guide on how to use the documentation tools as well as a reference of the various APIs. An additional benefit of this guide is that it would inform us of non backward-compatible changes.

nedbat:
@regisb We are definitely planning to thoroughly document the process of documenting APIs.  I guess @kdmccormick and I didn't let you know, but we have an ad-hoc to-do list in the old repo: https://github.com/edx/api-doc-tools/issues

I hear what you are saying about the api tooling perhaps being better in a separate repo.  I think the reasoning was that most repos that need edx-drf-extensions will also need the doc tooling, and vice-versa. Putting it all in one repo reduces the maintenance chores in the repos.  So it's a pragmatic choice rather than a philosophical one.  I could be swayed in the other direction.

regisb:
> I guess @kdmccormick and I didn't let you know, but we have an ad-hoc to-do list in the old repo: https://github.com/edx/api-doc-tools/issues

This is really cool! I wasn't aware that there was a dedicated repo.

> I think the reasoning was that most repos that need edx-drf-extensions will also need the doc tooling, and vice-versa.

The rub lies in the "most". If there is ever a non DRF-based API that comes out, it will have to install DRF nonetheless in order to document its API.

> Putting it all in one repo reduces the maintenance chores in the repos.

I disagree with this. For instance: if a repo depends on a feature included in release X+1 of api-doc-tools, but is not yet compatible with release X of edx-drf-extensions, then it will have to upgrade before it can start using the new feature.

> So it's a pragmatic choice rather than a philosophical one. I could be swayed in the other direction.

I'll do my best to convince you with practical arguments :)

kdmccormick:
@nedbat @regisb Thanks for all the feedback!

@nedbat I've applied some of your suggested changes; I'll apply the next week.

@regisb Re: `api-doc-tools` vs. `edx-drf-extensions`: You make good points. I hemmed and hawed between the two locations for a while. Within the walls of 141 Portland there seems to be a sentiment of "we have too many repositories", and every current edx.org IDA that needs API documentations also needs DRF extensions. However, I understand those two thing aren't necessarily true in the Open edX community. I'll stew on this and update you with my thoughts when I come back to this PR next week.

As a point of pedantry, the API tools are powered by `drf-yasg` which required DRF anyway.

Finally, regarding documentation on how to use these tools: that is absolutely part of the plan! My bad for omitting that from the description.

kdmccormick:
@regisb @nedbat Keeping this out of DRFX and making use of https://github.com/edx/api-doc-tools sounds good to me. I'll post a similar PR to that repo this week.

nedbat:
Weren't we going to finish the tweaks here before moving to the other repo?

kdmccormick:
@nedbat Right; that was the plan.

regisb:
> Keeping this out of DRFX and making use of https://github.com/edx/api-doc-tools sounds good to me. I'll post a similar PR to that repo this week.

Thanks! I understand the practical issues around the management of many different repos, and I appreciate the effort.

edx_api_doc_tools/tests/test_doc_tools.py
kdmccormick: @nedbat Realizing now that this is the maximal thing I can assert, and it's passing because that string is actually the `<title>`. The API spec is loaded via an AJAX call, which populates all page content.

edx_rest_framework_extensions/__init__.py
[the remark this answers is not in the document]
kdmccormick: (moot, because we're moving over to edx-api-doc-tools)

edx_api_doc_tools/conf_utils.py
kdmccormick: Added some examples.

kdmccormick:
@nedbat @regisb I finished addressing the comments on this PR, and will be working off of https://github.com/edx/api-doc-tools/pull/13 from now on. I'll ping you both when I'm ready for another round of review.

requirements/constraints.txt
```diff
@@ -8,5 +8,8 @@
 # pin when possible.  Writing an issue against the offending project and
 # linking to it here is good.
 
-# This package is a backport from python 3 that can only be installed in python 2.7
+# These packages are backports from python 3 that can only be installed in python 2.7
+backports.functools-lru-cache ; python_version == "2.7"
+functools32 ; python_version == "2.7"
 futures ; python_version == "2.7"
+ruamel.ordereddict ; python_version == "2.7"
```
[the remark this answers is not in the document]
kdmccormick: @nedbat It's a transitive dependency through `ruamel.yaml`. It fails to install in Py3, which is why I added this constraint.

PR #92, opened 2019-11-02
ARCH-1269: remove jwt authentication from csrf endpoint 

robrap:
**remove jwt authentication from csrf endpoint**

The main purpose of this commit is to remove the IsAuthenticated
permission from the /csrf/api/v1/token endpoint. The permission was
never needed, and doesn't allow the endpoint to be used to retrieve a
csrf token for use with an anonymous POST.

A temporary Django Setting was added to help with the roll-out:
- EDX_DRF_EXTENSIONS[ENABLE_ANONYMOUS_ACCESS_ROLLOUT]
This setting will help manage the rollout of CSRF protection.

Other minor fixes included:
- Updated README to clarify the purpose of the repo.
- Added testing of the existing csrf app to tox.

ARCH-1269

### DIFF
--- a/csrf/api/v1/views.py
+++ b/csrf/api/v1/views.py
@@ -3,11 +3,8 @@ API for CSRF application.
 """
 
 from django.middleware.csrf import get_token
-from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.views import APIView
-
-from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
 
 
 class CsrfTokenView(APIView):
@@ -25,9 +22,6 @@ class CsrfTokenView(APIView):
             >>>     "csrfToken": "abcdefg1234567"
             >>> }
     """
-
-    authentication_classes = (JwtAuthentication,)
-    permission_classes = (IsAuthenticated,)
 
     def get(self, request):
         """
--- a/edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
+++ b/edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
@@ -6,16 +6,18 @@ import ddt
 import mock
 from django.contrib.auth import get_user_model
 from django.test import RequestFactory, TestCase, override_settings
-from rest_framework.exceptions import AuthenticationFailed
+from rest_framework.exceptions import AuthenticationFailed, PermissionDenied
 from rest_framework_jwt.authentication import JSONWebTokenAuthentication
 
 from edx_rest_framework_extensions.auth.jwt import authentication
 from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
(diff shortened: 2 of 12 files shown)


csrf/api/v1/views.py
```diff
@@ -27,8 +28,11 @@ class CsrfTokenView(APIView):
     """
 
     authentication_classes = (JwtAuthentication,)
-    permission_classes = (IsAuthenticated,)
```
nasthagiri: If not requiring `IsAuthenticated`, why is `JwtAuthentication` still listed as an authentication class?

csrf/api/v1/views.py
```diff
@@ -27,8 +28,11 @@ class CsrfTokenView(APIView):
...
 
+    # Use requires_csrf_token on views that need a correct csrf_token, but without the
+    # CSRF protection that csrf_protect enforces.
```
nasthagiri: Can you explain this statement?  I'm having trouble parsing this sentence.  Perhaps it's a tongue-twister with the c's and the s's, etc.

edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
```diff
@@ -174,18 +176,44 @@ def test_authenticate_credentials_no_usernames(self):
...
-            with mock.patch.object(Logger, 'debug') as logger:
-                self.assertRaises(
-                    Exception,
-                    JwtAuthentication().authenticate,
-                    request
-                )
-                self.assertTrue(logger.called)
+    @ddt.data(
+        ('get', True, True, _MOCK_ANONYMOUS_RETURN),
+        ('get', True, True, _MOCK_USER_AUTH_RETURN),
```
nasthagiri: The 2 above cases have the same input values, but different expected outputs. ?

nasthagiri: Ah. I see that the last value is the mocked return value instead of the expected result.
Suggestion: `expected_user_and_auth` -> `mocked_return_value_user_and_auth`

csrf/api/v1/views.py
robrap: I will test if this affects the get_token() call and remove or comment accordingly.

robrap: Seems to work without JwtAuthentication, so I removed this.

csrf/api/v1/views.py
robrap: Although comments made me think this was needed, it seems to work without this, so I removed this.

edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
robrap: Updated tests as discussed.  Also, removed 'get' tests, because I think that added unnecessary complication.

PR #95, opened 2019-11-07
ARCH-1269: minor fixes to recent releases

robrap:
- restore public api toggle constants
- add permission of AllowAny to the csrf endpoint

The toggle constants for oauth scopes are used outside of
edx-drf-extensions (in edx-platform), so these constants must survive
and stay separate. NOTE: This problem was introduced in 2.4.2 while
edx-platform remained on 2.4.0. This will be fixed in edx-platform
while upgrading.

This fixes a csrf endpoint bug introduced into the 2.4.3 release,
where an IDA that had a default permission of DjangoModelPermissions
would start failing on this endpoint.

ARCH-1269

### DIFF
--- a/edx_rest_framework_extensions/config.py
+++ b/edx_rest_framework_extensions/config.py
@@ -2,8 +2,20 @@
 Application configuration constants and code.
 """
 
-OAUTH_TOGGLE_NAMESPACE = 'oauth2'
-SWITCH_ENFORCE_JWT_SCOPES = '{}.enforce_jwt_scopes'.format(OAUTH_TOGGLE_NAMESPACE)
+# .. toggle_name: oauth2.enforce_jwt_scopes
+# .. toggle_implementation: WaffleSwitch
+# .. toggle_default: False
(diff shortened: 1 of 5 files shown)


edx_rest_framework_extensions/config.py
```diff
@@ -2,8 +2,20 @@
...
+# .. toggle_default: False
+# .. toggle_description: Enforces JWT Scopes for an IDA. See https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/oauth_dispatch/docs/decisions/0006-enforce-scopes-in-LMS-APIs.rst
+# .. toggle_category: authorization
+# .. toggle_use_cases: incremental_release
+# .. toggle_creation_date: 2018-06-28
+# .. toggle_expiration_date: 2020-12-31
+# .. toggle_warnings: Toggle may be referenced from multiple IDAs.
+# .. toggle_tickets: ARCH-154
+# .. toggle_status: supported
+OAUTH_TOGGLE_NAMESPACE = 'oauth2'  # IMPORTANT: Constant is part of the public api.  Do NOT rename.
```
robrap: Reviewer: Restoring namespace and toggle names for use in edx-platform with WaffleSwitch that requires both.

PR #98, opened 2019-11-26
-switched to pytest

Ayub-Khan:
-switched to pytest


Ayub-Khan:
Closing in fav of https://github.com/edx/edx-drf-extensions/pull/99

PR #100, opened 2019-12-02
Clean up pip-tools setup and stricten pylint

kdmccormick:
Broken out from https://github.com/edx/edx-drf-extensions/pull/91 in order to commit the merge-conflict-susceptible code sooner than later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ clean: ## remove intermediate files
 piptools: ## install pip-compile and pip-sync.
 	pip install -r requirements/pip-tools.txt
 
-upgrade-piptools: piptools ## upgrade pip-tools using pip-tools.
+upgrade-piptools: piptools # upgrade pip-tools using pip-tools.
 	pip-compile requirements/pip-tools.in --rebuild --upgrade -o requirements/pip-tools.txt
 
 upgrade: export CUSTOM_COMPILE_COMMAND=make upgrade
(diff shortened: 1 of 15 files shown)


Makefile
```diff
@@ -13,38 +14,51 @@ clean: ## remove intermediate files
...
 
-quality: ## run quality checks using tox
-	tox -e quality
-
-piptools:
+piptools: ## install pip-compile and pip-sync.
 	pip install -r requirements/pip-tools.txt
 
-_upgrade: piptools
+upgrade-piptools: piptools ## upgrade pip-tools using pip-tools.
```
nedbat: Will people use this target explicitly? It used to be hidden.

Makefile
```diff
@@ -13,38 +14,51 @@ clean: ## remove intermediate files
...
 	mv requirements/test.tmp requirements/test.txt
-
-	# Delete line "-e file:///local/path/to/edx-drf-extensions", which
-	# is a result of the "-e ." hack in test.in.
-	sed '/^-e /d' requirements/test.txt > requirements/test.tmp
-	sed '/^-e /d' requirements/dev.txt > requirements/dev.tmp
-	mv requirements/test.tmp requirements/test.txt
-	mv requirements/dev.tmp requirements/dev.txt
-
 	# Generate pins for pip-tools itself.
```
nedbat: Shouldn't this comment go too?

kdmccormick: Yes

Makefile
kdmccormick: The diff is obfuscating the story a bit: `upgrade-piptools` is a completely new target, whereas `_upgrade` was a hack to set `CUSTOM_COMPILE_COMMAND`, which I removed because I learned a better way to set it. It's a coincidence that they have the same prerequisite.

Anyway, I changed `#` to `##` so that `upgrade-piptools` recipe won't be exposed in `make help`.

kdmccormick:
@nedbat Ready for another pass when you have the chance.

PR #101, opened 2019-12-03
Switch from nose to pytest

kdmccormick:
While working on https://github.com/edx/edx-drf-extensions/pull/91, I got frustrated with the poor error messages and weaker CLI of nose as compared to pytest.

#99 moves us to pytest anyway, but I'm not sure when it will merge, so this PR pulls out the nose-to-pytest-specific changes so that they can merge on their own.

cc @Ayub-Khan , @jmbowman , @nedbat

### DIFF
--- a/test_settings.py
+++ b/test_settings.py
@@ -13,7 +13,6 @@ INSTALLED_APPS = (
     'csrf.apps.CsrfAppConfig',
     'django.contrib.auth',
     'django.contrib.contenttypes',
-    'django_nose',
     'edx_rest_framework_extensions',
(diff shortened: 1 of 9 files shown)


test_settings.py
```diff
@@ -13,7 +13,6 @@
     'csrf.apps.CsrfAppConfig',
     'django.contrib.auth',
     'django.contrib.contenttypes',
-    'django_nose',
```
jmbowman: You should remove the `TEST_RUNNER` setting while you're at it, since it references a class in this package.

PR #104, opened 2020-01-14
Add basic manage.py file to allow direct pytest usage

kdmccormick:
@jinder1s This should fix the `ImportError: No module named 'test_settings'` thing that happens when you try to invoke `pytest` directly.

This is just a copypasta from another edX library.


kdmccormick:
@jinder1s should we merge or close this?

PR #105, opened 2020-02-03
Added new relic metrics to bearerAuthentication

jinder1s:
We might soon be updating BearerAuthentiation(or removing). Adding these here to gauge where it is still actively used.

### DIFF
--- a/edx_rest_framework_extensions/auth/bearer/authentication.py
+++ b/edx_rest_framework_extensions/auth/bearer/authentication.py
@@ -4,6 +4,7 @@ import logging
 
 import requests
 from django.contrib.auth import get_user_model
+from edx_django_utils.monitoring import set_custom_metric
 from rest_framework import exceptions
 from rest_framework.authentication import BaseAuthentication, get_authorization_header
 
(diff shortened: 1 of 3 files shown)


edx_rest_framework_extensions/auth/bearer/authentication.py
```diff
@@ -36,13 +37,16 @@ def get_user_info_url(self):
         return get_setting('OAUTH2_USER_INFO_URL')
 
     def authenticate(self, request):
+        set_custom_metric("BearerAuthentication", "Failed")
```
robrap: Please add the `# default value` comment.

edx_rest_framework_extensions/auth/bearer/authentication.py
```diff
@@ -81,6 +85,7 @@ def authenticate_credentials(self, token):
         if not user.is_active:
             raise exceptions.AuthenticationFailed('User inactive or deleted.')
 
+        set_custom_metric("BearerAuthentication", "Success")
```
robrap: Can you move this into the calling method with the out calls to `set_custom_metric`?  Just use a variable to capture the return, then set the metric, and then return.  Thanks.

PR #107, opened 2020-02-24
BOM-1324: remove oauth2.enforce_jwt_scopes toggle

robrap:
The oauth2.enforce_jwt_scopes waffle switch was added temporarily for
the rollout of JWT scopes. This removes the toggle and replaces it with
logic equivalent to setting `oauth2.enforce_jwt_scopes` to True.

BREAKING CHANGE:

This removes a toggle that may or may not have been set in any
particular environment, and was defaulted to False.

Before taking this upgrade:
* Make sure your IDA includes `EnsureJWTAuthSettingsMiddleware` in its
declared `MIDDLEWARE` or `MIDDLEWARE_CLASSES`.
* Although you could first check and/or set the
`oauth2.enforce_jwt_scopes` waffle switch to True in all environments
for your IDA, this upgrade is unlikely to cause an issue. If you want to
play it safe, setting the switch first is how you do it, but then you
need remove the switch.

After taking this upgrade:
* Once the upgrade has been deployed and is stable, delete the
`oauth2.enforce_jwt_scopes` waffle switch from all environments for the
IDA with the upgrade.

### DIFF
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '2.4.6'  # pragma: no cover
+__version__ = '3.0.0'  # pragma: no cover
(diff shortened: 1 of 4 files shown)


robrap:
Yes @timmc-edx.  I will add the commit comment to the releases found here: https://github.com/edx/edx-drf-extensions/releases

PR #108, opened 2020-03-05
BOM-1045 First Part

Ayub-Khan:
-Updated travis and tox file
-Added Manifest file
-Updated base requirements
-Updated setup.py
-Updated Makefile
-Updated Requirements


requirements/base.txt
```diff
@@ -6,26 +6,27 @@
 #
 certifi==2019.11.28       # via requests
 chardet==3.0.4            # via requests
-django-waffle==0.18.0     # via edx-django-utils
-django==1.11.26           # via edx-django-utils, rest-condition
+django-waffle==0.20.0
```
jmbowman: django-waffle 0.19.0 dropped support for Django 2.0 and 2.1.

requirements/base.txt
```diff
@@ -6,26 +6,27 @@
...
+django==2.2.11
```
jmbowman: There should probably be a constraint keeping this < 2.0 until we start testing the newer versions in CI.

setup.py
```diff
@@ -1,10 +1,45 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+from __future__ import absolute_import, print_function
```
jmbowman: With Python 2.7 support being dropped, we shouldn't need this.

tox.ini
```diff
@@ -1,17 +1,21 @@
...
 deps =
+    -r{toxinidir}/requirements/test.txt
     django111: Django>=1.11,<2.0
     django20: Django>=2.0,<2.1
     django21: Django>=2.1,<2.2
     django22: Django>=2.2,<2.3
-    -r{toxinidir}/requirements/test.txt
+    drf37: djangorestframework<3.8
+    drf38: djangorestframework<3.9
+    drf39: djangorestframework<4.0
```
jmbowman: As written this would currently get the 3.11.0 release, not the latest 3.9.x

requirements/test.txt
```diff
@@ -4,72 +4,68 @@
...
 edx-opaque-keys==2.0.1
-enum34==1.1.6             # via astroid
 factory-boy==2.12.0
-faker==2.0.4              # via factory-boy
-funcsigs==1.0.2           # via mock, pytest
+faker==4.0.1              # via factory-boy
+filelock==3.0.12          # via virtualenv
 future==0.18.2
-futures==3.1.1 ; python_version == "2.7"
+futures==3.1.1
```
jmbowman: I thought this package was both broken and unnecessary under Python 3; can we just remove it from `tests.in`?

PR #110, opened 2020-03-06
BOM-1098 Removing ENABLE_ANONYMOUS_ACCESS_ROLLOUT flag

jinder1s:
ENABLE_ANONYMOUS_ACCESS_ROLLOUT flag was a temporarily used to facilitate rollout
of CSFR protection for MFEs. With that effort finished, the flag is no longer necessary
and is now being removed.

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
+++ b/edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
@@ -17,7 +17,6 @@ from edx_rest_framework_extensions.auth.jwt.tests.utils import (
     generate_jwt_token,
     generate_latest_version_payload,
 )
-from edx_rest_framework_extensions.config import ENABLE_ANONYMOUS_ACCESS_ROLLOUT
 from edx_rest_framework_extensions.tests import factories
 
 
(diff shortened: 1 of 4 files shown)


edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
```diff
@@ -179,45 +178,21 @@ def test_authenticate_credentials_no_usernames(self):
...
-        with mock.patch.object(JSONWebTokenAuthentication, 'authenticate', return_value=mocked_return_value_user_and_auth):  # noqa E501 line too long
-            with override_settings(EDX_DRF_EXTENSIONS={ENABLE_ANONYMOUS_ACCESS_ROLLOUT: enable_rollout}):
-                actual_user_and_auth = JwtAuthentication().authenticate(request)
-
-        self.assertEqual(mocked_return_value_user_and_auth, actual_user_and_auth)
-
     @ddt.data(
         # CSRF protected because using JWT cookies to successfully authenticate (similar to SessionAuthentication)
-        (True, True, _MOCK_USER_AUTH_RETURN),
+        (True, _MOCK_USER_AUTH_RETURN),
```
feanil: Do we need DDT if there is only one set of params here?  Can we just flatten this out?

PR #116, opened 2020-03-15
ARCHBOM-1036: fix for edx-drf-extensions 3.0.0

robrap:
In 3.0.0, the switch `oauth2.enforce_jwt_scopes` was removed, which
starts checking is_restricted in JWTs. This works fine for JWTs created
with the LMS, but uncovered a pre-existing bug that will only show
itself in the Ecommerce Service for certain JWTs which were meant to be
decoded with a custom jwt_decode_handler. In the Ecommerce Service only,
this custom jwt_decode_handler is set using the JWT_DECODE_HANDLER
setting.

This fix updates the JWT code to respect the JWT_DECODE_HANDLER setting
of JWT_AUTH, and uses the configured handler rather than assuming the
edx-drf-extensions version will always be used.

ARCHBOM-1036

### DIFF
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '3.0.0'  # pragma: no cover
+__version__ = '3.0.1'  # pragma: no cover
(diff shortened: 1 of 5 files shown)


robrap:
**Reviewer Note:** Is `release/3.x` a reasonable name for the release branch?  Do we have a naming convention for this?

PR #124, opened 2020-04-21
Updated the tox and travis configs -- BOM-1553

mraarif:
Relevant JIRA issue [here](https://openedx.atlassian.net/browse/BOM-1553).


awais786:
Add this constraint.
```
# Stay on an LTS release
django<2.3
```

.travis.yml
```diff
@@ -1,42 +1,31 @@
...
-      env: TOXENV=quality
-    - python: 3.5
-      env: TOXENV=docs
-
+  - python: 3.5
+    env: TOXENV=quality
+  - python: 3.5
+    env: TOXENV=docs
+  - python: 3.8
+    env: TOXENV=quality
```
morenol, line 16: We can remove this from matrix and instread add a new env with TOXENV=quality which should be equivalent to have these two workers for quality test here.

.travis.yml
```diff
@@ -1,42 +1,31 @@
...
+  - python: 3.8
+    env: TOXENV=docs
```
morenol, line 18: Same as above

.travis.yml
mraarif: that's actually a very good suggestion, thanks

tox.ini
```diff
@@ -1,36 +1,31 @@
 [tox]
-envlist = py35-django{111,20,21,22}-drf{39,latest},quality, docs
+envlist = py{35,38}-django{22,30}-drf{39,latest},quality, docs
```
morenol, line 2: Django 3.0 with Python3.5 is an invalid combination

PR #125, opened 2020-05-05
ARCHBOM-128: enhance authentication metrics

robrap:
BREAKING CHANGE: Changed names of NewRelic metric names and values.
  Has potential to break NewRelic dashboards or alerts, but should not
  have an affect on deployed code.

- Renamed 'request_auth_type' to 'request_auth_type_guess'. This makes
  it more clear that this metric could be wrong in certain cases.
- Renamed value `session-or-unknown` to `session-or-other`. This name
  makes it more clear that it is the method of authentication that is in
  question, not whether or not the user is authenticated.
- Added 'jwt-cookie' as new value for 'request_auth_type_guess'.
- Added new 'request_authenticated_user_found_in_middleware' metric.
  Helps identify for what middleware step the request user was set, if
  it was set. Example values: 'process_request', 'process_view',
  'process_response', or 'process_exception'.
- Fixed/Added setting of authentication metrics for exceptions as well.
- Fixed values of 'unauthenticated' and 'no-user' for
  'request_auth_type_guess' to be more accurate.

ARCHBOM-128

### DIFF
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '5.1.0'  # pragma: no cover
+__version__ = '6.0.0'  # pragma: no cover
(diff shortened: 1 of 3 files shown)


robrap:
@feanil: Thanks. This middleware was and still is using the old-style.  I implemented more of the functions as an implementation detail of the items listed, so I don't think there is anything to add.  Happy to follow-up with more details if you think that would be helpful, or if you want more explanation.

robrap:
FYI: I tested in edx-platform https://github.com/edx/edx-platform/pull/23904

PR #131, opened 2020-06-01
Upgrade Python requirements files

timmc-edx:


timmc-edx:
I assume because we just haven't gotten to it yet. I'm not sure what process was followed in originally setting up the upgrade jobs.

PR #136, opened 2020-07-16
WIP: return None on NotAuthenticated

iamsobanjaved:
https://openedx.atlassian.net/browse/BOM-1845

https://github.com/Styria-Digital/django-rest-framework-jwt/pull/27/files#diff-bc0105dbb10d8b2bcacf8d191d5d0195R105
This change is causing issue. Previously it returns None but now it throws exception.

@awais786  have created a PR there https://github.com/Styria-Digital/django-rest-framework-jwt/pull/65

### DIFF
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '6.1.0'  # pragma: no cover
+__version__ = '6.1.1'  # pragma: no cover
--- a/edx_rest_framework_extensions/auth/jwt/authentication.py
+++ b/edx_rest_framework_extensions/auth/jwt/authentication.py
@@ -59,6 +59,12 @@ class JwtAuthentication(JSONWebTokenAuthentication):
         return get_setting('JWT_PAYLOAD_MERGEABLE_USER_ATTRIBUTES')
 
     def authenticate(self, request):
+        # latest drf-jwt version throws error for any other value other than jwt. So returns None and pass it to other
+        # Authentication class
+        auth_header_value = request.environ.get('HTTP_AUTHORIZATION')
+        if auth_header_value and not auth_header_value.lower().startswith('jwt'):
+            return None
+
         try:
             user_and_auth = super(JwtAuthentication, self).authenticate(request)
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,16 @@ Unreleased
 
 *
 
+[6.1.1] - 2020-07-19
+--------------------
+
+Fixed
+~~~~~~~
+
+* Latest `drf-jwt` is throwing error in case of any other Authorization Header. Fixing that issue in `JwtAuthentication` class.
+
+
+
 [6.1.0] - 2020-06-26
 --------------------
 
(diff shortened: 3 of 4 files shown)


edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -78,6 +78,8 @@ def authenticate(self, request):
 
         except jwt.InvalidTokenError:
             raise exceptions.AuthenticationFailed()
+        except exceptions.NotAuthenticated:
```
awais786, line 81: @robrap https://github.com/Styria-Digital/django-rest-framework-jwt/pull/27/files#diff-bc0105dbb10d8b2bcacf8d191d5d0195R105
This line throws error. Previously it return None.

https://github.com/Styria-Digital/django-rest-framework-jwt/issues/57

awais786: Issue already reported. But update.

edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -59,6 +59,10 @@ def get_jwt_claim_mergeable_attributes(self):
         return get_setting('JWT_PAYLOAD_MERGEABLE_USER_ATTRIBUTES')
 
     def authenticate(self, request):
+        # latest version of drf-jwt does not work with other auth header.
+        if request.environ.get('HTTP_AUTHORIZATION', '').startswith('Bear'):
```
awais786, line 63: @robrap https://github.com/Styria-Digital/django-rest-framework-jwt/pull/27/files#diff-bc0105dbb10d8b2bcacf8d191d5d0195R105
This line throws error. Previously it return None.

https://github.com/Styria-Digital/django-rest-framework-jwt/issues/57

awais786:
https://github.com/edx/ecommerce/pull/3026/files build passes.

edx_rest_framework_extensions/auth/jwt/authentication.py
robrap: Would it makes sense to test for not JWT instead?

edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -78,6 +82,8 @@ def authenticate(self, request):
 
         except jwt.InvalidTokenError:
             raise exceptions.AuthenticationFailed()
+        except exceptions.NotAuthenticated:
```
robrap, line 85: This feels like the right direction for allowing other auth to have a try.

robrap:
One more note is that it would be good to have a test for these cases. Thanks.

edx_rest_framework_extensions/auth/jwt/authentication.py
awais786: I just have concern that if it added a loop.

robrap: Not sure about loops, but it seems like this isn’t needed for your current concern.

edx_rest_framework_extensions/__init__.py
```diff
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '6.1.0'  # pragma: no cover
+__version__ = '6.1.1'  # pragma: no cover
```
robrap, line 3: Please update CHANGELOG.rst.

PR #139, opened 2020-09-23
ARCHBOM-1495: rename custom metric to custom attribute

robrap:
**Description:**

Rename of "custom metric" to "custom attribute" was based
on a decision (ADR) captured in edx-django-utils:
https://github.com/edx/edx-django-utils/blob/master/edx_django_utils/monitoring/docs/decisions/0002-custom-monitoring-language.rst

Deprecated RequestMetricsMiddleware due to rename.  Use
RequestCustomAttributesMiddleware instead.

**JIRA:**

[ARCHBOM-1495](https://openedx.atlassian.net/browse/ARCHBOM-1495)

**Testing:**

Note: I tested locally in edx-platform with both the deprecated name and new name.  The [PR to rename in edx-platform ](https://github.com/edx/edx-platform/pull/25082) will be failing until this version has been released and edx-platform has been updated.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bump if needed
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_rest_framework_extensions/middleware.py
+++ b/edx_rest_framework_extensions/middleware.py
@@ -1,6 +1,8 @@
 """
 Middleware to ensure best practices of DRF and other endpoints.
 """
+import warnings
+
 from django.utils.deprecation import MiddlewareMixin
 from edx_django_utils import monitoring
 from edx_django_utils.cache import DEFAULT_REQUEST_CACHE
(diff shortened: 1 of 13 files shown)


edx_rest_framework_extensions/middleware.py
```diff
@@ -164,3 +166,13 @@ def _cache_if_authenticated_user_found_in_middleware(self, request, value):
...
             DEFAULT_REQUEST_CACHE.set(self.AUTHENTICATED_USER_FOUND_CACHE_KEY, value)
+
+
+class RequestMetricsMiddleware(RequestCustomAttributesMiddleware):
+    """
+    Deprecated class for handling middleware. Class has been renamed to RequestCustomAttributesMiddleware.
+    """
+    def __init__(self, *args, **kwargs):
+        super(RequestMetricsMiddleware, self).__init__(*args, **kwargs)
+        msg = "Use 'RequestCustomAttributesMiddleware' in place of 'RequestMetricsMiddleware'."
```
feanil, line 177: Can we print where this instantiation is being called from or will that come through some other way in the logs?

feanil (approved):
If we can improve the warning or have another way of knowing where the object is being instantiated, no change required, otherwise I suggest providing the calling location as a part of the log message.

edx_rest_framework_extensions/middleware.py
robrap: Simply searching for the name made it very easy to find, so I don't think we need to add anything else.  See https://github.com/edx/edx-platform/pull/25082/files

PR #171, opened 2021-04-21
Added middleware for globally exposing headers such as Date

inventhouse:
**Description:**

Added middleware for globally exposing headers such as Date; access to the Date header will be used on the client-side to maintain a correction offset between the server time and the browser time as requests are made.  This offset will allow frontends to display much more reliable durations and expirations without making extra requests.

**Additional Details**

This is an expansion / generalization of https://github.com/edx/edx-platform/pull/27221

**Author concerns:** 

- Please confirm that I bumped the version correctly.  This adds a feature, but there should be no breaking changes.

- I didn't see where I should update documentation aside from CHANGELOG and docstrings; [docs/middleware.rst](https://github.com/edx/edx-drf-extensions/blob/master/docs/middleware.rst) is basically just `automodule`.

**Reviewers:**
- [x] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [x] CI build is green
- [x] Version bump if needed
- [x] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


CHANGELOG.rst
```diff
@@ -12,6 +12,15 @@ Change Log
 Unreleased
 ----------
 
+[6.6.0] - 2021-04-21
+--------------------
+
+Added
+~~~~~
+
+* Added AccessControlExposeHeadersMiddleware, a new middleware for globally exposing headers such as Date; access to the Date header will be used on the client-side to maintain a correction offset between the server time and the browser time as requests are made.  This offset will allow frontends to display much more reliable durations and expirations without making extra requests.
```
robrap, line 21: A couple of high-level thoughts. We can discuss more next week.
1. We do a lot of header manipulation in nginx. Where should we be doing this? We may want an ADR someone that documents why we are choosing one over the other.
2. This is related to work done for https://openedx.atlassian.net/browse/MICROBA-585, which relates to https://openedx.atlassian.net/browse/ARCHBOM-1150. This adds this data to the login_refresh endpoint that is already available to the MFE, but no one has had a chance to write the MFE code to make use of it.
3. The one place where the server might need to adjust, rather than the client, is for cookie expiration times.  However, this would require the client sending its time to the server. I suspect that there are cookies that never get created by the browser because they are expired on arrival, but I have yet to be able to prove that. The new data available from MICROBA-585 was meant to be able to test that theory. Note: if we decided to compensate on cookie expiration time, our JWT tokens would still be secure, because the token would still have the server expiration time and the server would still only trust the token, not the cookie.

robrap (commented):
Thanks @inventhouse. I wanted to share some thoughts. Some of the info I share might not be what you want to rely on, but in theory, you could. Let's discuss.

CHANGELOG.rst
inventhouse: 1. I don't have a strong opinion about whether to expose the date via this middleware or in nginx.
2. A single data point available on the client-side is certainly better than no time reference, but part of my idea is that this could be ubiquitous and continuously updated / refined and doesn't need to be added to any API.  I have a couple specific APIs using the the `Date` header on the frontend and am working on generalizing the code and integrating it into `frontend-platform`
3. Interesting, but like you said, the cookie issue likely involves sending the time the other way.

robrap: Regarding point 1:
* Maybe check in with someone from SRE to see if they have an opinion?
* Adding to nginx may be a quick way to get this everywhere without having to add this middleware to the cookiecutter and all IDAs.

Points 2 and 3 were simply related informs, and I agree they don't really affect this PR. I should have called that out from the start.

inventhouse:
Closing old PR

PR #192, opened 2021-07-30
build: Implemented common-constraint local sync.

awais786:
Implemented common-constraint local sync. It gives us more control in case of different version implementation.

pip-tools gives error if two different versions of same package appears in make upgrade job. To fix this problem, added common constraint locally. It will always sync with web-url. With this approach this issue can be resolve.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,8 @@ $(COMMON_CONSTRAINTS_TXT):
 upgrade: export CUSTOM_COMPILE_COMMAND=make upgrade
 upgrade: $(COMMON_CONSTRAINTS_TXT)
 upgrade: upgrade-piptools piptools ## upgrade requirement pins.
+	sed 's/pyjwt\[crypto\]==1.7.1//g' requirements/common_constraints.txt > requirements/common_constraints.tmp
+	mv requirements/common_constraints.tmp requirements/common_constraints.txt
 	pip-compile --allow-unsafe --rebuild -o requirements/pip.txt requirements/pip.in
 	pip-compile requirements/base.in --upgrade -o requirements/base.txt
 	pip-compile requirements/test.in --upgrade -o requirements/test.txt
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -21,8 +21,16 @@ piptools: ## install pip-compile and pip-sync.
...
 
+COMMON_CONSTRAINTS_TXT=requirements/common_constraints.txt
+.PHONY: $(COMMON_CONSTRAINTS_TXT)
+$(COMMON_CONSTRAINTS_TXT):
+	wget -O "$(@)" https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt || touch "$(@)"
+
 upgrade: export CUSTOM_COMPILE_COMMAND=make upgrade
+upgrade: $(COMMON_CONSTRAINTS_TXT)
 upgrade: upgrade-piptools piptools ## upgrade requirement pins.
+	sed 's/pyjwt\[crypto\]==1.7.1//g' requirements/common_constraints.txt > requirements/common_constraints.tmp
```
awais786, line 32: Removing constraint from here. 
URL brings constraint `1.7.1` if i tried to override in locally it gives error.

requirements/constraints.txt
```diff
@@ -9,7 +9,7 @@
 # linking to it here is good.
 
 # This file contains all common constraints for edx-repos
--c https://raw.githubusercontent.com/edx/edx-lint/master/edx_lint/files/common_constraints.txt
+-c common_constraints.txt
 
 # greater versions causing tests failures.
 pyjwt[crypto]==1.7.1
```
awais786, line 15: This will remove in next PR ( upgrade `pyjwt[crypto]` )

mraarif (approved):

UsamaSadiq (approved):

PR #252, opened 2022-05-31
docs: generalize "edx-search" to "search" in pagination docstring

kdmccormick:
## Description

When this code was extracted from edx-platform,
it was likely used to paginate search results
specifically from the "edx-search" library
(http://github.com/openedx/edx-search/).

Now that this is generic utility function,
it makes more sense to just say
"search results" instead of
"edx-search results", as the
result list can come from any backend.

## Reviewers

@georgebabey , as the original author, do you mind reviewing this docs tweak?


kdmccormick:
Merging without requirements because this is such a trivial change.

PR #254, opened 2022-06-08
fix: Installed pip and pip-tools in upgrade script

Jawayria:
Updated the upgrade target script to check the compatibility of upgraded pip and pip-tools versions.
For reference, look at this https://github.com/openedx/edx-repo-health/pull/271 for new check added in edx-repo-health.

JIRA: https://2u-internal.atlassian.net/browse/BOM-3455


Makefile
```diff
@@ -32,6 +32,7 @@ upgrade: upgrade-piptools piptools ## upgrade requirement pins.
 	sed 's/pyjwt\[crypto\]<2.0.0//g' requirements/common_constraints.txt > requirements/common_constraints.tmp
 	mv requirements/common_constraints.tmp requirements/common_constraints.txt
 	pip-compile --allow-unsafe --rebuild -o requirements/pip.txt requirements/pip.in
+	pip install -qr requirements/pip.txt
```
UsamaSadiq, line 35: We should move all commands inside `make upgrade` target to keep the changes consistent among all repos.

PR #256, opened 2022-06-15
feat: enhance get_decoded_jwt

robrap:
**Description:**

The get_decoded_jwt will now also call get_decoded_jwt_from_auth,
so it doesn't need to be called separately and can use the cached
Jwt.

Also added monitoring custom attributes to determine if it is
worthwhile to add additional caching.

**JIRA:**

2U Private: ARCHBOM-2129

**Reviewers:**
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bump if needed
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/cookies.py
+++ b/edx_rest_framework_extensions/auth/jwt/cookies.py
@@ -2,9 +2,22 @@
 JWT Authentication cookie utilities.
 """
 
-from django.conf import settings
+import logging
 
+from django.conf import settings
+from edx_django_utils.monitoring import increment as increment_custom_attribute
+from edx_django_utils.monitoring import set_custom_attribute
+from rest_framework.request import Request as DrfRequest
+from rest_framework.settings import api_settings
+
+from edx_rest_framework_extensions.auth.jwt.authentication import (
+    JwtAuthentication,
+    get_decoded_jwt_from_auth,
+)
 from edx_rest_framework_extensions.auth.jwt.decoder import configured_jwt_decode_handler
+
+
+logger = logging.getLogger(__name__)
 
 
 def jwt_cookie_name():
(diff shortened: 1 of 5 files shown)


robrap:
Decided this is not needed. Closing.

edx_rest_framework_extensions/auth/jwt/cookies.py
```diff
@@ -23,11 +36,47 @@ def get_decoded_jwt(request):
...
+        drf_request = request
+        if not isinstance(request, DrfRequest):
+            drf_request = DrfRequest(request, parsers=api_settings.DEFAULT_PARSER_CLASSES)
+        user_and_auth = JwtAuthentication().authenticate(drf_request)
+        if user_and_auth is not None:
+            return configured_jwt_decode_handler(user_and_auth[1])
+
+        logger.info('get_decoded_jwt: Failed to decode Jwt cookie.')
+    except Exception as ex:  # pylint: disable=broad-except
+        logger.info(f'get_decoded_jwt: Unknown error decoding Jwt cookie. {ex!r}')
```
robrap, line 80: @feanil: In digging up PRs from the time that mobile switched to JWTs, I found this one.
1. I'm wondering if these "info" logs should really be debug logs.
2. I'm wondering it this could use your new error handling code. Or, even simpler, if this were switched to debug we could provide the full exception.

feanil: I think using the new error extraction code here would be useful and then I would leave them as `INFO` since the debug logs don't get published by default in production.

robrap: @feanil: Oh geez. I didn't realize I decided to close this PR instead of merging it. And, I don't remember what I was trying to do to begin with. It possibly had something to do with the JWT mobile work, but I decided at the time this wasn't needed, so let's just trust earlier me. :)

PR #268, opened 2022-08-16
feat: Error message when JWT user disabled

moeez96:
**Description:**

If user has been disabled by admin, JWT Token Authentication should fail with appropriate message.
Disabled user refers to user's password being set as unusable by admin on page `/support/manage_user`

**JIRA:**

[LEARNER-9001](https://2u-internal.atlassian.net/browse/LEARNER-9001)

**Additional Details**

* **Dependencies:**: None
* **Merge deadline:** August 19th 2022
* **Author concerns:** None

**Testing instructions:**
* Fetch JWT token for an edx user.
* Call the api `/api/mobile/v1/users/honor/course_enrollments/` with the JWT token.
* See that response is received.
* Disable the user from the page `/support/manage_user`.
* Call api again with same user.
* Observe authentication failed error `User disabled by admin: <username>`

**Reviewers:**
- [@jawad-khan  ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bump if needed
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '8.2.0'  # pragma: no cover
+__version__ = '8.3.0'  # pragma: no cover
--- a/edx_rest_framework_extensions/auth/jwt/authentication.py
+++ b/edx_rest_framework_extensions/auth/jwt/authentication.py
@@ -14,6 +14,9 @@ from edx_rest_framework_extensions.auth.jwt.decoder import configured_jwt_decode
 from edx_rest_framework_extensions.settings import get_setting
 
 
+JWT_USER_DISABLED_ERROR = 'user_is_disabled'
+
+
 logger = logging.getLogger(__name__)
 
 
@@ -66,6 +69,18 @@ class JwtAuthentication(JSONWebTokenAuthentication):
             # Unauthenticated, CSRF validation not required
             if not user_and_auth:
                 return user_and_auth
+
+            # Fail authentication if user disabled
+            try:
+                user = user_and_auth[0]
+            except IndexError:
+                user = None
+            if user:
+                if not user.has_usable_password():
+                    msg = 'User disabled by admin: %s' % user.get_username()
+                    raise exceptions.AuthenticationFailed({
+                        'error_code': JWT_USER_DISABLED_ERROR,
+                        'developer_message': msg})
 
             # Not using JWT cookies, CSRF validation not required
             use_jwt_cookie_requested = request.META.get(USE_JWT_COOKIE_HEADER)
--- a/edx_rest_framework_extensions/auth/jwt/authentication.py
+++ b/edx_rest_framework_extensions/auth/jwt/authentication.py
@@ -75,9 +75,7 @@ class JwtAuthentication(JSONWebTokenAuthentication):
             if user and isinstance(user, get_user_model()):
                 if not user.has_usable_password():
                     msg = 'User disabled by admin: %s' % user.get_username()
-                    raise exceptions.AuthenticationFailed({
-                        'error_code': JWT_USER_DISABLED_ERROR,
-                        'developer_message': msg})
+                    raise exceptions.AuthenticationFailed(code=JWT_USER_DISABLED_ERROR, detail=msg)
 
             # Not using JWT cookies, CSRF validation not required
             use_jwt_cookie_requested = request.META.get(USE_JWT_COOKIE_HEADER)
--- a/edx_rest_framework_extensions/auth/jwt/authentication.py
+++ b/edx_rest_framework_extensions/auth/jwt/authentication.py
@@ -71,11 +71,8 @@ class JwtAuthentication(JSONWebTokenAuthentication):
                 return user_and_auth
 
             # Fail authentication if user disabled
-            try:
-                user = user_and_auth[0]
-            except IndexError:
-                user = None
-            if user:
+            user = user_and_auth[0]
+            if user and isinstance(user, get_user_model()):
                 if not user.has_usable_password():
                     msg = 'User disabled by admin: %s' % user.get_username()
                     raise exceptions.AuthenticationFailed({
--- a/edx_rest_framework_extensions/auth/jwt/authentication.py
+++ b/edx_rest_framework_extensions/auth/jwt/authentication.py
@@ -71,9 +71,10 @@ class JwtAuthentication(JSONWebTokenAuthentication):
             user = user_and_auth[0]
             if user and isinstance(user, get_user_model()):
                 if not user.has_usable_password():
-                    msg = 'User is disabled.'
-                    logger.exception(msg)
-                    raise exceptions.AuthenticationFailed(msg)
+                    log_message = 'User id {} attempted JWT authentication after being disabled by ' \
+                                  'an admin.'.format(user.id)
+                    logger.exception(log_message)
+                    raise exceptions.AuthenticationFailed('User is disabled.')
 
             # Not using JWT cookies, CSRF validation not required
             use_jwt_cookie_requested = request.META.get(USE_JWT_COOKIE_HEADER)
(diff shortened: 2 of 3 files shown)


edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -67,6 +70,18 @@ def authenticate(self, request):
             if not user_and_auth:
                 return user_and_auth
 
+            # Fail authentication if user disabled
```
jawad-khan, line 73: No need for this try, if array is empty then function will return in above if.

edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -67,6 +70,18 @@ def authenticate(self, request):
...
+            try:
+                user = user_and_auth[0]
+            except IndexError:
+                user = None
+            if user:
```
jawad-khan, line 78: Add unit test for this exception.

edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
```diff
@@ -177,11 +177,16 @@ def test_authenticate_credentials_no_usernames(self):
     @mock.patch('edx_rest_framework_extensions.auth.jwt.authentication.set_custom_attribute')
     def test_authenticate_csrf_protected(self, mock_set_custom_attribute):
         """ Verify authenticate exception for CSRF protected cases. """
+
+        username = 'ckramer'
```
jawad-khan, line 181: why do we need this?

edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
```diff
@@ -191,6 +196,20 @@ def test_authenticate_csrf_protected(self, mock_set_custom_attribute):
...
+        request = RequestFactory().get('/', HTTP_AUTHORIZATION=jwt_token)
+
+        username = 'ckramer'
+        email = 'ckramer@hotmail.com'
+        user = factories.UserFactory(email=email, username=username, is_staff=False)
+
+        with mock.patch.object(JSONWebTokenAuthentication, 'authenticate', return_value=(user, "mock-auth")):
+            with mock.patch.object(User, 'has_usable_password', return_value=False):
+                with self.assertRaises(AuthenticationFailed):
+                    JwtAuthentication().authenticate(request)
```
jawad-khan, line 211: There are lot of AuthenticationFailed error raising from this class, can we please check the error code?

edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -67,6 +70,15 @@ def authenticate(self, request):
             if not user_and_auth:
                 return user_and_auth
 
+            # Fail authentication if user disabled
+            user = user_and_auth[0]
+            if user and isinstance(user, get_user_model()):
+                if not user.has_usable_password():
+                    msg = 'User disabled by admin: %s' % user.get_username()
+                    raise exceptions.AuthenticationFailed({
```
jawad-khan, line 78: can we please follow the error msg standard already present in JWT implementation?

edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
moeez96: We need to create a user to mock the response in: https://github.com/openedx/edx-drf-extensions/blob/a4c55012de9e7d762cf6dd0fe74f4bdb34107895/edx_rest_framework_extensions/auth/jwt/authentication.py#L67 The expected response is a tuple containing a user instance and data.
 It was previously mocked incorrectly as a tuple of strings.

edx_rest_framework_extensions/auth/jwt/authentication.py
moeez96: @jawad-khan Can you please specify? In the current file I do not see any message being passed in `AuthenticationFailed()`.  
On the other hand, the current message added is in line with the message passed in bearer token authentication in the same case (User disabled by admin).

robrap:
Closing this PR because of external discussion where we agreed that JWTs will be valid until they expire.

robrap:
If we reopen this, we should confirm there is no extra db call to determine if disabled. I’m guessing that is loaded with the user, but please verify. Thank you.

moeez96:
@robrap It seems like no extra db call is made to fetch user's disability status since Django's code [here ](https://github.com/django/django/blob/286e7d076cfb7b4b0bdfc917de3020e6e89683f6/django/contrib/auth/hashers.py#L35)only compares the user's current password hash to a hardcoded string.

Shall we go ahead and reopen this?

edx_rest_framework_extensions/__init__.py
```diff
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '8.1.0'  # pragma: no cover
+__version__ = '8.1.1'  # pragma: no cover
```
robrap, line 3: Please make this 8.2.0 (minor bump in version).

edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -67,6 +70,13 @@ def authenticate(self, request):
             if not user_and_auth:
                 return user_and_auth
 
+            # Fail authentication if user disabled
+            user = user_and_auth[0]
+            if user and isinstance(user, get_user_model()):
+                if not user.has_usable_password():
+                    msg = 'User disabled by admin: %s' % user.get_username()
+                    raise exceptions.AuthenticationFailed(code=JWT_USER_DISABLED_ERROR, detail=msg)
```
robrap, line 78: Will we do the same change in error and log info we did elsewhere?

robrap (commented):
@moeez96: If you have confirmed no extra db call, the we'll just want to watch this carefully when it goes to Production to ensure it doesn't protect against anything that is legitimate by mistake.

edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
moeez96: Done

edx_rest_framework_extensions/auth/jwt/authentication.py
moeez96: Done

edx_rest_framework_extensions/__init__.py
moeez96: bumped to 8.3.0 since 8.2.0 was found on master after rebasing.

moeez96:
Reopened PR, rebased from master branch and implemented feedback.

robrap (commented):
1. Thanks. Just minor comments.
2. Can you plan to upgrade in edx-platform right after this gets released, so we can ensure all goes smoothly. I still have a hesitancy, although it should be fine.

edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -67,6 +67,14 @@ def authenticate(self, request):
             if not user_and_auth:
                 return user_and_auth
 
+            # Fail authentication if user disabled
+            user = user_and_auth[0]
+            if user and isinstance(user, get_user_model()):
+                if not user.has_usable_password():
+                    msg = 'User is disabled.'
+                    logger.exception(msg)
```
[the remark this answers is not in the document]
moeez96: This would be an exception because the user is being denied entry. I believe the info may be used to log data while user is following expected flow of use.

moeez96:
@robrap 
3. I believe disabled user instances matching this criteria can only be created via edx support tool (for disabling user) or the django shell. In either case, I believe other IDA's do not support this flow/method of enabling/disabling a user. Unless it is django based and users are explicitly disabled by admins/SRE's, in which case the expected is them to be blocked on services.
4. Yes, I will final squash for the message to be "feat:".
5. Rebased branch with master. Please review

edx_rest_framework_extensions/auth/jwt/authentication.py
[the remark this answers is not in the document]
robrap: 1. Would you mind making the log message more descriptive? Something like:
```
User id XXX attempted JWT authentication after being disabled by an admin.
```
2. I think an error log would be fine, since we don't really need the stacktrace. Hopefully my last note on this. :)

robrap (commented):
Apologies. I was confusing myself with `has_usable_password`, thinking it might be related to None or empty string, but the newer version is to return False only when `set_unusable_password` is set. I know you know this, but somehow it didn't sink in for me.

One minor comment about the log message, and this should be good to go.

robrap (approved):
Thanks!

PR #270, opened 2022-08-23
feat: Raise exception decoding symmetric jwt if flag is false

jawad-khan:
Raise exception decoding symmetric jwt if decode_symmetric_token is false

**Description:**

By default jwt_decode_handler will decode symmetric tokens but if we pass decode_symmetric_token as False,
it will raise InvalidTokenError with NoSuitableSigningKeys in __cause__ attr of the exxception. Caller can determine whether the
exception was because of NoSuitableSigningKeys or something else went wrong.

**JIRA:**

[LEARNER-8941](https://2u-internal.atlassian.net/browse/LEARNER-8941)

**Additional Details**

* **Dependencies:**: None.

**Reviewers:**
- [x] tag reviewer

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bump if needed
- [x] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,14 @@ Change Log
 Unreleased
 ----------
 
+[8.1.1] - 2022-08-24
+--------------------
+
+Added
+~~~~~
+
+* Added only asymmetric jwt decoding functionality in decoder
+
 * Rename toggle_warnings to toggle_warning for consistency with setting_warning.
 
 [8.1.0] - 2022-01-28
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,13 +12,16 @@ Change Log
 Unreleased
 ----------
 
-[8.1.1] - 2022-08-24
+[8.2.0] - 2022-08-24
 --------------------
 
 Added
 ~~~~~
 
 * Added only asymmetric jwt decoding functionality in decoder
+
+Changed
+~~~~~~~
 
 * Rename toggle_warnings to toggle_warning for consistency with setting_warning.
 
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '8.1.0'  # pragma: no cover
+__version__ = '8.1.1'  # pragma: no cover
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '8.1.1'  # pragma: no cover
+__version__ = '8.2.0'  # pragma: no cover
--- a/edx_rest_framework_extensions/auth/jwt/decoder.py
+++ b/edx_rest_framework_extensions/auth/jwt/decoder.py
@@ -35,6 +35,8 @@ def jwt_decode_handler(token, decode_symmetric_token=True):
     Notes:
         * Requires "exp" and "iat" claims to be present in the token's payload.
         * Aids debugging by logging InvalidTokenError log entries when decoding fails.
+        * Setting for JWT_DECODE_HANDLER expects a single argument, token. The new argument i.e. decode_symmetric_token
+          is for internal use only.
 
     Examples:
         Use with `djangorestframework-jwt <https://getblimp.github.io/django-rest-framework-jwt/>`_, by changing
@@ -71,12 +73,26 @@ def jwt_decode_handler(token, decode_symmetric_token=True):
     return _set_token_defaults(decoded_token)
 
 
-def configured_jwt_decode_handler(token, decode_symmetric_token=True):
+def configured_jwt_decode_handler(token):
     """
     Calls the ``jwt_decode_handler`` configured in the ``JWT_DECODE_HANDLER`` setting.
     """
     api_setting_jwt_decode_handler = api_settings.JWT_DECODE_HANDLER
-    return api_setting_jwt_decode_handler(token, decode_symmetric_token)
+    return api_setting_jwt_decode_handler(token)
+
+
+def get_asymmetric_only_jwt_decode_handler(token):
+    """
+    Returns a jwt_decode_handler that will only validate asymmetrically signed JWTs.
+
+    WARNING: This will only work with a service that is configured to use the
+       jwt_decode_handler from this library. This can be used to decode an
+       already decoded JWT, to ensure it is asymmetrically signed. This check
+       can go away once the DEPR for symmetrically signed JWTs is complete:
+       https://github.com/openedx/public-engineering/issues/83
+    """
+    api_setting_jwt_decode_handler = api_settings.JWT_DECODE_HANDLER
+    return api_setting_jwt_decode_handler(token, decode_symmetric_token=False)
 
 
 def decode_jwt_scopes(token):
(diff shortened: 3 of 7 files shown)


edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -65,17 +66,17 @@ def jwt_decode_handler(token):
...
 
 
-def configured_jwt_decode_handler(token):
+def configured_jwt_decode_handler(token, decode_symmetric_token=True):
     """
     Calls the ``jwt_decode_handler`` configured in the ``JWT_DECODE_HANDLER`` setting.
     """
     api_setting_jwt_decode_handler = api_settings.JWT_DECODE_HANDLER
-    return api_setting_jwt_decode_handler(token)
+    return api_setting_jwt_decode_handler(token, decode_symmetric_token)
```
robrap, line 79: Hmm. I don't think we can just change the signature of this setting that expects a handler that just takes a token. However, I (or you) will need to think about what this means.

For example, here is the ecommerce "configured" handler: https://github.com/GetBlimp/django-rest-framework-jwt

We may need a separate method that doesn't get the configured handler, but just uses the drf-extensions handler to do some post-processing? These are just thoughts off the top of my head so we can think further about this.

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -65,17 +66,17 @@ def jwt_decode_handler(token):
...
     """
     jwt_issuer = get_first_jwt_issuer()
-    _verify_jwt_signature(token, jwt_issuer)
+    _verify_jwt_signature(token, jwt_issuer, decode_symmetric_token=decode_symmetric_token)
     decoded_token = _decode_and_verify_token(token, jwt_issuer)
     return _set_token_defaults(decoded_token)
 
 
-def configured_jwt_decode_handler(token):
+def configured_jwt_decode_handler(token, decode_symmetric_token=True):
```
robrap, line 74: Since `configured_jwt_decode_handler` has a setting and expected signature with one argument, token, I recommend we leave this function as-is.

Instead, let's add a new function:
```
get asymmetric_only_jwt_decode_handler(token):
   """
   Returns a jwt_decode_handler that will only validate asymmetrically signed JWTs.

   WARNING: This will only work with a service that is configured to use the
       jwt_decode_handler from this library. This can be used to decode an
       already decoded JWT, to ensure it is asymmetrically signed. This check
       can go away once the DEPR for symmetrically signed JWTs is complete:
       https://github.com/openedx/public-engineering/issues/83
   """
```
This call can be used on the `request.auth` validated JWT token from LMS.

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -56,6 +56,7 @@ def jwt_decode_handler(token):
 
     Args:
         token (str): JWT to be decoded.
+        decode_symmetric_token (bool): Whether to decode symmetric tokens or not. Pass False for asymmetric tokens only
```
robrap, line 59: We should note that the setting for JWT_DECODE_HANDLER expects a single argument, token. The new argument is for internal use only.

edx_rest_framework_extensions/__init__.py
```diff
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '8.1.0'  # pragma: no cover
+__version__ = '8.1.1'  # pragma: no cover
```
robrap, line 3: Please make this (and changelog) 8.2.0. Thank you.

CHANGELOG.rst
```diff
@@ -12,6 +12,14 @@ Change Log
...
 
+[8.1.1] - 2022-08-24
+--------------------
+
+Added
+~~~~~
+
+* Added only asymmetric jwt decoding functionality in decoder
+
 * Rename toggle_warnings to toggle_warning for consistency with setting_warning.
```
robrap, line 23: Please add a `Changed` header above this bullet.

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -28,13 +28,15 @@ class JwtTokenVersion:
...
 
-def jwt_decode_handler(token):
+def jwt_decode_handler(token, decode_symmetric_token=True):
     """
     Decodes (and verifies) a JSON Web Token (JWT).
 
     Notes:
         * Requires "exp" and "iat" claims to be present in the token's payload.
         * Aids debugging by logging InvalidTokenError log entries when decoding fails.
+        * Setting for JWT_DECODE_HANDLER expects a single argument, token. The new argument i.e. decode_symmetric_token
```
robrap, line 38: Minor changes so the comment remains true even when this argument is no longer _new_.
```suggestion
        * Setting for JWT_DECODE_HANDLER expects a single argument, token. The argument decode_symmetric_token
```

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -78,6 +81,20 @@ def configured_jwt_decode_handler(token):
...
+    Returns a jwt_decode_handler that will only validate asymmetrically signed JWTs.
+
+    WARNING: This will only work with a service that is configured to use the
+       jwt_decode_handler from this library. This can be used to decode an
+       already decoded JWT, to ensure it is asymmetrically signed. This check
+       can go away once the DEPR for symmetrically signed JWTs is complete:
+       https://github.com/openedx/public-engineering/issues/83
+    """
+    api_setting_jwt_decode_handler = api_settings.JWT_DECODE_HANDLER
+    return api_setting_jwt_decode_handler(token, decode_symmetric_token=False)
```
robrap, line 95: Here we won't use the decoder from the setting.
```suggestion
    return jwt_decode_handler(token, decode_symmetric_token=False)
```

edx_rest_framework_extensions/auth/jwt/tests/test_decoder.py
```diff
@@ -177,6 +178,32 @@ def test_required_claims(self, claim):
...
+        Verifies the function logs decode failures with symmetric token set as false,
+        and raises an InvalidTokenError if token is symmetric
+        """
+        # Create valid token symmetric jwt token and set decode_symmetric_token as False.
+        with mock.patch('edx_rest_framework_extensions.auth.jwt.decoder.logger') as patched_log:
+            with self.assertRaises(jwt.InvalidTokenError):
+                token = generate_jwt_token(self.payload)
+                jwt_decode_handler(token, decode_symmetric_token=False)
+
+            patched_log.exception.assert_any_call("Token verification failed.")
```
robrap, line 192: I don't think we need this. I'd say we just test this via the other new test. This is just my opinion though. Feel free to disagree.

edx_rest_framework_extensions/auth/jwt/tests/test_decoder.py
```diff
@@ -177,6 +178,32 @@ def test_required_claims(self, claim):
...
+    def test_failure_decode_asymmetric(self):
+        """
+        Verifies the function logs decode failures when provided a symmetric token to asymmetric decode method,
+        and raises an InvalidTokenError if token is symmetric
+        """
+        # Create valid token symmetric jwt token and use asymmetric method to decode.
+        with mock.patch('edx_rest_framework_extensions.auth.jwt.decoder.logger') as patched_log:
+            with self.assertRaises(jwt.InvalidTokenError):
+                token = generate_jwt_token(self.payload)
+                get_asymmetric_only_jwt_decode_handler(token)
```
robrap, line 203: Can we get a success test for this new function?

robrap (commented):
Thanks @jawad-khan.

robrap (approved):
Looks good. Thanks. Ready to squash, merge and publish.

PR #287, opened 2022-12-16
feat: add enduser.id custom attribute for monitoring

robrap:
**Description:**

Added custom attribute enduser.id, following OpenTelemetry convention. The old custom attribute request_user_id should be considered deprecated. This works with some New Relic automatic tooling around users. See https://docs.newrelic.com/docs/errors-inbox/error-users-impacted/

Task: https://github.com/edx/edx-arch-experiments/issues/140

**Merge checklist:**
- [X] Version bump if needed
- [X] Changelog record added
- [X] Documentation updated (not only docstrings)
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_rest_framework_extensions/tests/test_middleware.py
+++ b/edx_rest_framework_extensions/tests/test_middleware.py
@@ -116,20 +116,26 @@ class TestRequestCustomAttributesMiddleware(TestCase):
         self.request.user = UserFactory()
 
         self.middleware.process_response(self.request, None)
-        mock_set_custom_attribute.assert_any_call('request_user_id', self.request.user.id)
-        mock_set_custom_attribute.assert_any_call(
-            'request_authenticated_user_found_in_middleware', 'process_response'
-        )
+
+        expected_calls = [
+            call('enduser.id', self.request.user.id),
+            call('request_user_id', self.request.user.id),
+            call('request_authenticated_user_found_in_middleware', 'process_response'),
+        ]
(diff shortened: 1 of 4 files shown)


edx_rest_framework_extensions/tests/test_middleware.py
```diff
@@ -116,20 +116,26 @@ def test_request_user_id_attribute(self, mock_set_custom_attribute):
...
 
         self.middleware.process_response(self.request, None)
-        mock_set_custom_attribute.assert_any_call('request_user_id', self.request.user.id)
-        mock_set_custom_attribute.assert_any_call(
-            'request_authenticated_user_found_in_middleware', 'process_response'
-        )
+
+        expected_calls = [
+            call('enduser.id', self.request.user.id),
+            call('request_user_id', self.request.user.id),
```
rgraber, line 122: Can we just remove this now that it's deprecated?

rgraber (commented):
One request to remove a test since I don't think we want to test a deprecated field

edx_rest_framework_extensions/tests/test_middleware.py
robrap: 1. The old if could be used in alerts or dashboards, so it should remain tested.
2. I did just rename the test function to match the non-deprecated attribute name.

rgraber (approved):
LGTM

PR #304, opened 2023-02-16
feat: replace pyjwkest with authlib

mumarkhan999:
TBD

PR #300, opened 2023-02-09
replace pyjwkest with cryptojwt

mumarkhan999:
Related PR: https://github.com/openedx/edx-platform/pull/31756

Issue #290, opened 2023-01-09
Replace pyjwkest


jmbowman:
The pyjwkest package was abandoned upstream more than 3 years ago.  Let's replace it with something more actively maintained.  First check to see if the latest version of PyJWT (already used here) supports what we need from pyjwkest.  If not, I nominate [Authlib](https://pypi.org/project/Authlib/) as seeming to cover the same functionality with much more usage and active maintenance, but feel free to do a quick search on PyPI or elsewhere for a better replacement.

jmbowman:
Maybe https://github.com/jpadilla/pyjwt/pull/723 added what we need?  Not sure without further digging into how this works.

robrap:
Please see https://github.com/openedx/edx-drf-extensions/issues/333 for related discussion. I do not know if we have custom decoding code for asymmetric JWTs, or only for the symmetric JWTs.

PR #311, opened 2023-03-16
feat: replace pyjwkest with pyjwt

mumarkhan999:
As we're removing `pyjwkest` from our dependencies, this PR is in the continuation of that [task](https://github.com/openedx/edx-drf-extensions/issues/290).

In this PR the `JWK` private-public keypair has been updated as the new package `PyJWT` which we're using instead of `Pyjwkest` is not compatible with the older ones.

---

**Similar PRs**
- https://github.com/openedx/edx-platform/pull/31829
- https://github.com/openedx/auth-backends/pull/233

**Another ticket**
- https://github.com/openedx/edx-platform/issues/31927

**Slack thread [Private]**
- 🧵 https://twou.slack.com/archives/C04ACDVM6A1/p1678302149454029

**Tool used for JWK keypair creation**
- 🛠 https://github.com/mumarkhan999/rsa-key-generator

**Previous attempts**
- Using [authlib](https://github.com/lepture/authlib)
  - https://github.com/openedx/edx-drf-extensions/pull/304
- Using [cryptojwt](https://github.com/IdentityPython/JWTConnect-Python-CryptoJWT)
  - https://github.com/openedx/edx-drf-extensions/pull/300


edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -173,9 +173,20 @@ def _set_filters(token):
...
     key_set = _get_signing_jwk_key_set(jwt_issuer, add_symmetric_keys=decode_symmetric_token)
-
     try:
-        _ = JWS().verify_compact(token, key_set)
+        for i in range(0, len(key_set)):
+            try:
+                _ = jwt.decode(
+                        token,
+                        key=key_set[i].key,
+                        algorithms=['HS256', 'RS512'],
```
timmc-edx, line 182: I think this is still vulnerable. One of these is a symmetric algorithm and the other is asymmetric. If you allow both, that means the attacker can take your RS512 _public_ key, then use it in HS256 as the _secret_ key, which will then allow verification to succeed. We need to ensure that the keys we supply here match the intended use of each key.

More generally, we should be relying on a JWT library to do this kind of looping and key-matching operation, rather than rewriting it ourselves, since it's a very delicate and error-prone part of JWT cryptography.

test_settings.py
```diff
@@ -46,24 +46,29 @@
     'JWT_SECRET_KEY': 'test-key',
 
     'JWT_PUBLIC_SIGNING_JWK_SET': (
-        '{"keys": [{"kid": "BTZ9HA6K", "e": "AQAB", "kty": "RSA", "n": "o5cn3ljSRi6FaDEKTn0PS-oL9EFyv1pI7dRgffQLD1qf5D6'
-        'sprmYfWWokSsrWig8u2y0HChSygR6Jn5KXBqQn6FpM0dDJLnWQDRXHLl3Ey1iPYgDSmOIsIGrV9ZyNCQwk03wAgWbfdBTig3QSDYD-sTNOs3pc'
-        '4UD_PqAvU2nz_1SS2ZiOwOn5F6gulE1L0iE3KEUEvOIagfHNVhz0oxa_VRZILkzV-zr6R_TW1m97h4H8jXl_VJyQGyhMGGypuDrQ9_vaY_RLEu'
-        'lLCyY0INglHWQ7pckxBtI5q55-Vio2wgewe2_qYcGsnBGaDNbySAsvYcWRrqDiFyzrJYivodqTQ"}]}'
+        '{"keys": [{"use": "sig", "e": "AQAB", "kty": "RSA", "alg": "RS512", "n": "kuI1T8prOHqX7Q9ARgyi7DsyLGMEGOmhPefU'
```
timmc-edx, line 49: [request] Can you please describe what the role of these changes is? Do we need to switch from RSA-256 to 512 in order to make this upgrade? Is the `use` field now mandatory, or is this an opportunistic improvement of some kind? etc.

edx_rest_framework_extensions/auth/jwt/decoder.py
timmc-edx: I believe that in the standard JWT algorithm-confusion attack, the attacker needs to acquire the RSA public key and use it to sign via the HS256 algorithm. So you'd have to do something more custom on the token generation side in order to demonstrate this. More details: https://portswigger.net/web-security/jwt/algorithm-confusion

Oof, that code you linked to in pyjwkest sure doesn't inspire confidence. It looks like it trusts the `"alg"` field in the untrusted header, unless the JWK has an `"alg"` field of its own. :-/

Anyway, I think there's a not-too-terrible solution: Change `_get_signing_jwk_key_set` to return a list of (alg, key) tuples so that the algorithm always matches the key. (It could check if the key has an alg already defined, and if not, fall back to a hardcoded RS256 or HS256).  This is potentially a breaking change, since some deployers may not have an alg field defined in their keys, but I think for edx-platform we'll need to do a key rotation anyhow.

timmc-edx: (Ah, pyjwkest has code that narrows the keyset to just the keys with a matching algorithm -- see `alg2keytype` and `pick_keys`.)

test_settings.py
mumarkhan999: ### 1

We're not switching from `RSA-256` to `512` as we're already using `RSA-512` 

https://github.com/edx/edx-internal/search?q=%22ALGORITHM%22

### 2
These fields are added by the tool which I used to generate the test key-value pair of the key.
Note: Even if we remove these additional fields, the tests still pass

timmc-edx: 1. Ah OK! Somewhere I'd gotten the impression that the current ones were RS256. (Probably I was confusing key sizes and signature sizes.)
2. If new keys are needed, we'll want to create the actual replacement keys for our deployed environments in a secure manner. Can you document a way of generating a new keypair, so an engineer can do it on their laptop? We'll need to verify that the new key works with the existing library, too; I can do that by starting a key rotation in the staging environment. We can then use the same method to create the test key here, so that we know it matches in format.

It would also be great to use a multiline string with the JWT displayed in indented, key-sorted JSON so that it will be easier to read the code (and any diffs, in the future.)

mumarkhan999: I've updated the keys which I've generated using this script https://github.com/mumarkhan999/rsa-key-generator.
I've also tested these keys on the master branch as well. https://github.com/openedx/edx-drf-extensions/pull/315.
The keys are in intended and sorted JSON now.

mumarkhan999: @timmc-edx I've tested the new format of keys on the following sandbox.
- https://newkey.sandbox.edx.org/
I updated the `yml configuration` files from the sandbox server to add the new keys. Things are working fine.

timmc-edx: OK, great! I ended up reworking that script a bit as https://github.com/openedx/edx-platform/pull/32025/files although, as the comment on that PR points out, this duplicates an existing management command. So that PR might not actually merge. :-)

I'll use that script to generate keys and rotate them in stage, prod, and edge (https://github.com/edx/edx-arch-experiments/issues/261) and after that we'll be able to merge these changes.

timmc-edx: (This might take a bit -- I'm working with SRE to clarify the instructions for rotating these keys safely.)

timmc-edx:
Could you change the commit and PR commit type to `feat:`? `chore:` is really just for completely mechanical changes that are "uninteresting", like routine dependency upgrades that are thought not to have any effect, or translation updates, etc. Since these commit types are used in generating changelogs, it's better to err on the side of feat/fix -- chore gets hidden. (This goes for the other PRs in this group as well.)

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -173,9 +173,26 @@ def _set_filters(token):
 
 def _verify_jwt_signature(token, jwt_issuer, decode_symmetric_token):
```
timmc-edx, line 174: On master, Robert has made some changes to this method that duplicate the `JMS().verify_compact` call to add instrumentation around symmetric and asymmetric key usage. I believe he had intended to have that instrumentation removed by the time this PR was ready, but this one might be ready first.

My recommendation would be to move the for-block into its own private method (maybe `_verify_jwt_signature_inner`? no strong preference) or even just a closure so that that single call can be duplicated. Then you'll be able to merge or rebase to pick up the new changes.

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -173,9 +173,26 @@ def _set_filters(token):
...
+                    algorithms = ['RS256', 'RS512',]
+                elif key_set[i].key_type == 'oct':
+                    algorithms = ['HS256',]
+
+                _ = jwt.decode(
+                        token,
+                        key=key_set[i].key,
+                        algorithms=algorithms,
+                        audience=jwt_issuer['AUDIENCE'],
+                        options={'verify_signature': True}
```
timmc-edx, line 190: We can leave off this option, as `verify_signature` is True by default.

test_settings.py
timmc-edx: We've updated the existing keys in our deployed environments (all except sandboxes, which work differently) using a script I added in https://github.com/openedx/edx-platform/pull/32089 and I've added that to the Release Notes. Since we're going to be upgrading existing keys rather than replacing them, I've made a PR against your branch that does the same to the existing test keys: https://github.com/openedx/edx-drf-extensions/pull/337

edx_rest_framework_extensions/auth/jwt/decoder.py
timmc-edx: Please also remember to bump the version and add a changelog entry.

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -299,15 +320,16 @@ def _get_signing_jwk_key_set(jwt_issuer, add_symmetric_keys=True):
...
     # asymmetric keys
     signing_jwk_set = settings.JWT_AUTH.get('JWT_PUBLIC_SIGNING_JWK_SET')
     if signing_jwk_set:
-        key_set.load_jwks(signing_jwk_set)
+        key_set.extend(PyJWKSet.from_json(signing_jwk_set).keys)
 
     if add_symmetric_keys:
-        # symmetric key
-        key_set.add({'key': jwt_issuer['SECRET_KEY'], 'kty': 'oct'})
+        # symmetric_key
```
timmc-edx, line 331: [nit] Looks like an accidental change here of a space to an underscore.

timmc-edx (approved):

timmc-edx (approved):

timmc-edx:
Looks like you dropped the 8.8.0 version bump again. I'll add that in a new PR and do a release.

PR #353, opened 2023-06-16
fix: Added support for Django 4.2.

zubairshakoorarbisoft:
Added django42 upgrade support, to have more context on it please refer to the following [issue](https://github.com/edx/upgrades/issues/140)


edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -17,6 +18,11 @@
 
 
 class CSRFCheck(CsrfViewMiddleware):
+
+    def __init__(self):
```
awais786, line 22: Please check this.

zubairshakoorarbisoft: What should I check specifically? I think there is not Any dummy response issue.

.github/workflows/ci.yml
```diff
@@ -14,7 +14,13 @@ jobs:
       matrix:
         os: [ubuntu-20.04]
         python-version: ['3.8']
-        toxenv: [quality, docs,  django32-drf312, django32-drflatest]
+        toxenv: [
+          quality,
+          docs,
+          django32-drf314,
```
UsamaSadiq, line 20: `drf-314` is the `drf-latest` so we can remove this one as it'll be duplicate with the below envs.
```suggestion
```

UsamaSadiq (approved):

PR #354, opened 2023-06-16
feat: make jwt decode fucntion generic

mumarkhan999:
**Related ticket**
- https://github.com/openedx/edx-drf-extensions/issues/290

**Related PR**
- https://github.com/openedx/edx-platform/pull/32270

**Tasks**
- [X] Make the verify signature functionality generic 
- [X] Add flags for key set so that symmetric and asymmetric can be added as per demand

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/decoder.py
+++ b/edx_rest_framework_extensions/auth/jwt/decoder.py
@@ -193,8 +193,8 @@ def _verify_jwt_signature(token, jwt_issuer, decode_symmetric_token):
     #   can fully retire code paths for symmetric keys, as part of
     #   DEPR: Symmetric JWTs: https://github.com/openedx/public-engineering/issues/83
 
-    # Use add_symmetric_keys=False to only include asymmetric keys at first
-    key_set = _get_signing_jwk_key_set(jwt_issuer, add_symmetric_keys=False)
+    # Pass only asymmetric_keys to only include asymmetric keys at first
+    key_set = get_verification_jwk_key_set(asymmetric_keys=settings.JWT_AUTH.get('JWT_PUBLIC_SIGNING_JWK_SET'))
     # .. custom_attribute_name: jwt_auth_verify_asymmetric_keys_count
     # .. custom_attribute_description: Number of JWT verification keys in use for this
     #   verification. Should be same as number of asymmetric public keys. This is
@@ -203,7 +203,7 @@ def _verify_jwt_signature(token, jwt_issuer, decode_symmetric_token):
     set_custom_attribute('jwt_auth_verify_asymmetric_keys_count', len(key_set))
 
     try:
-        _verify_jwk_signature_using_keyset(token, key_set, jwt_issuer)
+        verify_jwk_signature_using_keyset(token, key_set, aud=jwt_issuer['AUDIENCE'])
         # .. custom_attribute_name: jwt_auth_asymmetric_verified
         # .. custom_attribute_description: Whether the JWT was successfully verified
         #   using an asymmetric key.
@@ -218,7 +218,9 @@ def _verify_jwt_signature(token, jwt_issuer, decode_symmetric_token):
     #   the asymmetric keys here is redundant and unnecessary, but this code is temporary and
     #   will be simplified once symmetric keys have been fully retired.
 
-    key_set = _get_signing_jwk_key_set(jwt_issuer, add_symmetric_keys=decode_symmetric_token)
+    asymmetric_keys = settings.JWT_AUTH.get('JWT_PUBLIC_SIGNING_JWK_SET')
+    secret_key = jwt_issuer['SECRET_KEY'] if decode_symmetric_token else None
+    key_set = get_verification_jwk_key_set(asymmetric_keys=asymmetric_keys, secret_key=secret_key)
     # .. custom_attribute_name: jwt_auth_verify_all_keys_count
     # .. custom_attribute_description: Number of JWT verification keys in use for this
     #   verification. Should be same as number of asymmetric public keys, plus one if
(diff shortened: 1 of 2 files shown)


edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -315,21 +327,21 @@ def _decode_and_verify_token(token, jwt_issuer):
...
     for verifying signatures.
     """
     key_set = []
 
-    # asymmetric keys
-    signing_jwk_set = settings.JWT_AUTH.get('JWT_PUBLIC_SIGNING_JWK_SET')
-    if signing_jwk_set:
+    if add_asymmetric_keys:
+        # asymmetric keys
+        signing_jwk_set = settings.JWT_AUTH.get('JWT_PUBLIC_SIGNING_JWK_SET')
```
timmc-edx, line 339: [concern] I think this would be a breaking change in behavior, possibly an unwanted one -- if `JWT_AUTH.JWT_PUBLIC_SIGNING_JWK_SET` is not defined, then from_json would throw.

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -315,21 +327,21 @@ def _decode_and_verify_token(token, jwt_issuer):
     return decoded_token
 
 
-def _get_signing_jwk_key_set(jwt_issuer, add_symmetric_keys=True):
+def get_signing_jwk_key_set(secret_key, add_symmetric_keys=True, add_asymmetric_keys=True):
```
timmc-edx, line 330: [discuss] It feels odd to specify the secret key in the args, but not the asymmetric keys. It also feels odd to specify both the secret key and whether to use it at all.

What if the signature were instead `(asymmetric_keys, secret_key)`? The caller would just get the appropriate settings. If a value is None, it is not used. (I also would like to change the name before making it part of the public API—these are not signing keys, but verification keys.)

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -248,7 +248,16 @@ def _verify_jwt_signature(token, jwt_issuer, decode_symmetric_token):
         raise jwt.InvalidTokenError(exc_info[2]) from token_error
 
 
-def _verify_jwk_signature_using_keyset(token, key_set, jwt_issuer):
+def verify_jwk_signature_using_keyset(token, key_set, aud=None, iss=None, verify_signature=True, verify_exp=True):
+
```
timmc-edx, line 252: This should get a docstring if it's public. (PR should also include a changelog.)

timmc-edx:
I see that you requested another review, but I still have some outstanding comments. Not sure if you saw those?

mumarkhan999:
> I see that you requested another review, but I still have some outstanding comments. Not sure if you saw those?

[Inform] Apologies for the late reply. I was busy with pymemcache and dot related tasks.

Yeah. I think I overlooked one comment at that time.

edx_rest_framework_extensions/auth/jwt/decoder.py
mumarkhan999: Added.

edx_rest_framework_extensions/auth/jwt/decoder.py
mumarkhan999: Updated.

edx_rest_framework_extensions/auth/jwt/decoder.py
mumarkhan999: Updated.

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -315,21 +361,19 @@ def _decode_and_verify_token(token, jwt_issuer):
     return decoded_token
 
 
-def _get_signing_jwk_key_set(jwt_issuer, add_symmetric_keys=True):
+def get_verification_jwk_key_set(asymmetric_keys=None, secret_key=None):
     """
     Returns a JWK Keyset containing all active keys that are configured
     for verifying signatures.
```
timmc-edx, line 367: Docstring should be updated -- currently describes reading from configuration. Maybe something like this:

```suggestion
    Creates a JWK Keyset containing the provided keys.
    
    Args:
        asymmetric_keys (list or None): List of asymmetric JWK verification keys,
            each in JSON format.
        secret_key (str or None): Secret key for symmetric JWT verification, as an
            unencoded string.
```

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -218,7 +218,9 @@ def _verify_jwt_signature(token, jwt_issuer, decode_symmetric_token):
     #   the asymmetric keys here is redundant and unnecessary, but this code is temporary and
     #   will be simplified once symmetric keys have been fully retired.
 
-    key_set = _get_signing_jwk_key_set(jwt_issuer, add_symmetric_keys=decode_symmetric_token)
+    asymmetric_keys = settings.JWT_AUTH.get('JWT_PUBLIC_SIGNING_JWK_SET')
```
timmc-edx, line 221: [nit, optional] Consider raising this to around line 196, where it can be used by the other verification call.

timmc-edx (commented):
Some minor things remaining, one just a suggestion.

edx_rest_framework_extensions/auth/jwt/decoder.py
mumarkhan999: done.

edx_rest_framework_extensions/auth/jwt/decoder.py
mumarkhan999: done.

timmc-edx (approved):

PR #366, opened 2023-07-26
refactor: Recombine original/forgiving

timmc-edx:
The fundamental change here is to the guard on setting request.user; we have to decide whether to pay attention to USE_JWT_COOKIE vs. the mere presence of the JWT cookies.

- Early exit in missing USE_JWT_HEADER scenario (when not forgiving)
- Drop the `request_jwt_cookie` attribute (we think it's unused)
- Move setting of request.user to end of method
- Remove indirection for set-request-user


robrap (approved):
Thanks for pairing on this.

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -230,50 +230,10 @@ def process_view(self, request, view_func, view_args, view_kwargs):  # pylint: d
...
-                log_message, attribute_value = self._get_missing_cookie_message_and_attribute(
-                    jwt_cookie_header_payload_name()
-                )
-            if not signature_cookie:
-                log_message, attribute_value = self._get_missing_cookie_message_and_attribute(
-                    jwt_cookie_signature_name()
-                )
-            log.warning(log_message)
-        else:
-            attribute_value = 'missing-both'
```
timmc-edx, line 272: It looks like we actually have one dashboard still referencing this: https://one.newrelic.com/redirect/entity/ODgxNzh8VklafERBU0hCT0FSRHw5MzY0MjI

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -230,50 +230,10 @@ def process_view(self, request, view_func, view_args, view_kwargs):  # pylint: d
...
-            log.warning('Both JWT auth cookies missing. JWT auth cookies will not be reconstituted.')
-        monitoring.set_custom_attribute('request_jwt_cookie', attribute_value)
+        if not is_forgiving_jwt_cookies_enabled:
+            if not request.META.get(USE_JWT_COOKIE_HEADER):
```
robrap, line 234: **Note to self:** Create a temp custom attribute for the old use-JWT-header, that appears whether or not forgiving JWTs are enabled. This will allow us to see endpoints that are now affected by JWT cookies now, but were not earlier.

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: [inform] I added a note to the dashboard so it could be updated once the new attributes are available.

edx_rest_framework_extensions/auth/jwt/middleware.py
robrap: Done.

robrap:
@timmc-edx: This is ready for you to re-review. Please review one commit at a time. I did make one simplification that changes the original (when deciding whether to set request user). As I explain in that commit, I'd be fine with the change even if we weren't doing forgiving JWTs and wouldn't feel the need to put it behind a toggle. Also, if I happen to be wrong for some reason, it would be useful to know this separate from enabling forgiving JWTs.

Once you approve, I can inform Feanil, if he is interested in taking a quick look before merging. Thanks!

feanil:
@robrap It looks like the setting of `request.user` here makes use of the cached user if it's available.  At a quick glance I don't see any middleware above this middleware that sets `request.user` [in edx-platform](https://github.com/openedx/edx-platform/blob/master/lms/envs/common.py#L2123)  but it looks like the UserStandingMiddleware which is pretty close after the `JwtAuthCookieMiddleware` is making use of `request.user` so it may break if we don't set it here.

robrap:
@feanil: A couple of clarifications:
1. Explictly what changed is that in the legacy case (toggle disabled), rather than setting request user for all scenarios where `USE_JWT_COOKIE_HEADER` is supplied, including when there is no JWT cookie, we only set the request user if the JWT cookie has been reconstituted.
  - Since the header is only supplied by MFEs, this would only affect MFE calls with no cookies, or the exceptional case where only one of the two cookies was supplied. 
  - Additionally, since for the most part MFEs are calling `/refresh_token` to ensure that the cookies are updated before making backend calls, this would be a rare scenario.
  - Lastly, as written, the code for when the toggle enabled will only set the request user when the cookie is available.
2. This doesn't actually affect edx-platform at this time. The toggle ENABLE_SET_REQUEST_USER_FOR_JWT_COOKIE is not on for edx-platform (and possibly most services), because it caused a memory leak in edx-platform. See [setting annotation notes](https://github.com/openedx/edx-drf-extensions/blob/7702a9fca7c4a6364882e5735b647ef46660e714/edx_rest_framework_extensions/config.py#L5-L14).

feanil:
Ah, thanks for the clarification.  I was concerned because [this call](https://github.com/openedx/edx-drf-extensions/blob/aa649e9ab13273a0624bfb623d4c80112631b601/edx_rest_framework_extensions/auth/jwt/middleware.py#L259-L261) won't happen but if you're not worried about that, it sounds like this is fine.

timmc-edx (commented):
Looks good to me!

GitHub won't let me approve, since I made the PR. You'll have to approve it instead. :-P

feanil (approved):

PR #375, opened 2023-08-30
fix: exceptional case with JwtAuthentication

robrap:
**Description:**

Fixes exceptional case where JwtAuthentication should not CSRF protect a request that has both a JWT token in the authorization header and a JWT cookie, since the cookie should be ignored.

Note: I had minor concerns that Mobile could have issues here, and I just didn't want to have that concern, so I took care of the exceptional case.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bump if needed
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
+++ b/edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
@@ -285,7 +285,7 @@ class JwtAuthenticationTests(TestCase):
         auth_header = '{token_name} {token}'.format(token_name='Bearer', token='abc123')
         request = RequestFactory().get('/', HTTP_AUTHORIZATION=auth_header)
         self.assertIsNone(JwtAuthentication().authenticate(request))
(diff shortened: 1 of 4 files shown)


edx_rest_framework_extensions/auth/jwt/tests/test_authentication.py
```diff
@@ -266,6 +266,19 @@ def test_authenticate_with_incorrect_jwt_authorization(self, mock_set_custom_att
             JwtAuthentication().authenticate(request)
         mock_set_custom_attribute.assert_any_call('jwt_auth_result', 'failed-auth-header')
 
+    @mock.patch('edx_rest_framework_extensions.auth.jwt.authentication.set_custom_attribute')
+    def test_authenticate_with_correct_jwt_authorization_and_bad_cookie(self, mock_set_custom_attribute):
```
robrap, line 270: **Note to reviewer:** I confirmed that this test fails without my change.

PR #397, opened 2023-10-23
docs: Update the security e-mail address.

feanil:
This repository is now managed by the Axim Collaborative and security issues
with it should be reported to security@openedx.org instead of security@edx.org

This work is being done as a part of https://github.com/openedx/wg-security/issues/16


nedbat (approved):

feanil:
Codecov is not sending an update and everything else has passed.  Given that the change is a doc change I'm gonna use my admin powers to merge this.

PR #398, opened 2023-10-25
fix: update jwt vs session user monitoring

robrap:
**Description:**

An issue was found when both ENABLE_JWT_VS_SESSION_USER_CHECK and ENABLE_SET_REQUEST_USER_FOR_JWT_COOKIE were enabled at the same time in an ecommerce stage environment for edx.org, that is not reproducible locally. Additionally, the issue killed the requests without providing errors, so potentially caused an infinite loop or some such issue.

The guess is that the code for setting the user behind ENABLE_SET_REQUEST_USER_FOR_JWT_COOKIE, and the code to get the user in the new ENABLE_JWT_VS_SESSION_USER_CHECK check, were somehow in conflict. This update attempts to skip the JWT vs Session check in the case that we have set the user based on the JWT, in which case it would be a JWT vs JWT check, which not only is unnecessary, but also may be the cause of the issue.

Adds custom attributes:
- set_user_from_jwt_status
- skip_jwt_vs_session_check

**Issue:**

https://github.com/edx/edx-arch-experiments/issues/429

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bump if needed
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/middleware.py
+++ b/edx_rest_framework_extensions/auth/jwt/middleware.py
@@ -5,15 +5,19 @@ import logging
 
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.middleware import get_user
+from django.contrib.auth.models import AnonymousUser
 from django.utils.deprecation import MiddlewareMixin
 from django.utils.functional import SimpleLazyObject
-from edx_django_utils import monitoring
 from edx_django_utils.cache import RequestCache
+from edx_django_utils.monitoring import set_custom_attribute
 from rest_framework.permissions import OperandHolder, SingleOperandHolder
 from rest_framework.request import Request
 from rest_framework.settings import api_settings
 from rest_framework_jwt.authentication import JSONWebTokenAuthentication
 
+from edx_rest_framework_extensions.auth.jwt.authentication import (
+    set_flag_is_request_user_set_for_jwt_auth,
+)
 from edx_rest_framework_extensions.auth.jwt.constants import (
     JWT_DELIMITER,
     USE_JWT_COOKIE_HEADER,
@@ -210,7 +214,6 @@ class JwtAuthCookieMiddleware(MiddlewareMixin):
         )
 
     """
-
     def _get_missing_cookie_message(self, cookie_name):
         """ Returns missing cookie log_message """
         return '{} cookie is missing. JWT auth cookies will not be reconstituted.'.format(
@@ -230,7 +233,7 @@ class JwtAuthCookieMiddleware(MiddlewareMixin):
         # .. custom_attribute_name: use_jwt_cookie_requested
         # .. custom_attribute_description: True if USE_JWT_COOKIE_HEADER was found.
         #   This is a temporary attribute, because this header is being deprecated/removed.
-        monitoring.set_custom_attribute('use_jwt_cookie_requested', bool(request.META.get(USE_JWT_COOKIE_HEADER)))
+        set_custom_attribute('use_jwt_cookie_requested', bool(request.META.get(USE_JWT_COOKIE_HEADER)))
 
         if not get_setting(ENABLE_FORGIVING_JWT_COOKIES):
             if not request.META.get(USE_JWT_COOKIE_HEADER):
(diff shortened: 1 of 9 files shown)


edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -258,19 +261,56 @@ def process_view(self, request, view_func, view_args, view_kwargs):  # pylint: d
...
+    return RequestCache(__name__).data
+
+
+def _get_cached_user_from_jwt(request, view_func):
+    """
+    Returns cached user from JWT authentication.
+
+    Performs JWT authentication if not already cached.
+    """
+    # Similar to django/contrib/auth/middleware.py get_user.
```
robrap, line 296: **Note to reviewer:** I know this layer of caching may not be necessary, since we don't yet have it. But it seemed like the implementation should be more parallel to what exists in the Django middleware.

@feanil: Feel free to comment or mark resolved if you have no comment.

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -258,19 +261,56 @@ def process_view(self, request, view_func, view_args, view_kwargs):  # pylint: d
...
         if has_reconstituted_jwt_cookie and get_setting(ENABLE_SET_REQUEST_USER_FOR_JWT_COOKIE):
-            # DRF does not set request.user until process_response. This makes it available in process_view.
-            # For more info, see https://github.com/jpadilla/django-rest-framework-jwt/issues/45#issuecomment-74996698
-            request.user = SimpleLazyObject(lambda: _get_user_from_jwt(request, view_func))
+            # Since this call to the user is not made lazily, and has the potential to cause issues, we
+            # ensure it is only used in the case of ENABLE_SET_REQUEST_USER_FOR_JWT_COOKIE.
+            if not get_user(request).is_authenticated:
+                # Similar to django/contrib/auth/middleware.py AuthenticationMiddleware.
+                set_flag_is_request_user_set_for_jwt_auth()
+                request.user = SimpleLazyObject(lambda: _get_cached_user_from_jwt(request, view_func))
```
robrap, line 283: **Note to reviewer:** A flag was used, rather than looking for `SimpleLazyObject`, because Django's AuthenticationMiddleware also uses `SimpleLazyObject`. The flag makes things 100% explicit, so I decided on this approach.

@feanil: Feel free to comment or mark as resolved.

robrap (commented):
@feanil: Here is a first pass that has existing tests passing (locally at least). I did not yet add new tests for the new functionality. I wanted some feedback before either of us invested in that next level of effort. Thanks.

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -258,19 +261,56 @@ def process_view(self, request, view_func, view_args, view_kwargs):  # pylint: d
...
+
+        # DRF authentication does not set the request.user early enough for it to be used in process_request/
+        # process_view of middleware. This code enables JWT cookie authentication to set the request.user for
+        # middleware, before it will presumably happen again during DRF authentication.
+        # For more info on DRF and the request.user, see
+        # https://github.com/jpadilla/django-rest-framework-jwt/issues/45#issuecomment-74996698
+        #
+        # If the user has already been authenticated, we already have a user and intentionally avoid resetting it. As of
+        # Oct 2023, this would likely be due to session authentication in middleware. It is thus important for
+        # JwtAuthentication to verify that the session user and JWT user match. It is possible that this would be better
```
robrap, line 274: @feanil: I also was also wondering about this case. In JwtAuthentication, if `ENABLE_SET_REQUEST_USER_FOR_JWT_COOKIE` **and** we get a JWT vs Session user mismatch, it might makes sense to error out, since we know it is possible that the middleware analytics and the JWT authentication may be for different users. Here we may want to fail even when JwtAuthentication succeeds, which we aren't yet doing for any other cases. Thoughts?

robrap: **Update**: I added enforcement for this in a new commit that involved the least amount of refactoring I could think of.

feanil: Sounds good, I agree that we should error out on mismatch here as well.

edx_rest_framework_extensions/auth/jwt/middleware.py
feanil: Cool, that makes sense, I like the explicitness.

edx_rest_framework_extensions/auth/jwt/middleware.py
feanil: This makes sense, I'll resolve after the review.

feanil (approved):

robrap:
@feanil: This is ready for re-review. You can review the three newest commits one-by-one. I made some minor changes and added tests. Thanks.

feanil (approved):
Changes look good.

robrap:
@feanil: This is ready for re-re-review. I want to see how many approvals I can get for the same PR. :) Please see the latest 3 commits, and reviewing them separately may make things simpler for you.

Issue #403, opened 2023-11-08
Publish and link to readthedocs documentation


robrap:
The README does not contain a link to the documentation. (The README might also not match the current recommendation in other ways.) I presume these docs are not yet published on readthedocs, but maybe they are and we are simply missing a link?

nedbat:
Let me know if you need help puzzling through readthedocs, etc.

PR #404, opened 2023-11-08
feat!: make forgiving JWTs the default

robrap:
**Description:**

This is the final step (in this library) of the rollout of forgiving JWTs as a replacement for the USE-JWT-COOKIE header.

**BREAKING CHANGE:** Removed ENABLE_FORGIVING_JWT_COOKIES toggle. It is now permanently enabled.
- The header USE-JWT-COOKIE was removed because it has been fully replaced by forgiving JWTs.
- Removed temporary rollout custom attributes: use_jwt_cookie_requested, jwt_auth_request_user_not_found, and skip_jwt_vs_session_check.

See ADR 0002-remove-use-jwt-cookie-header.rst for details.

**Reviewers:**
- [ ] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bump if needed
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.


docs/decisions/0002-remove-use-jwt-cookie-header.rst
robrap, line 1: **Note to reviewer:** 
- Ideally this file name would be changed, but I didn't want to deal with conflicts and that would be its own change, and I think I am just going to skip this. It is also referenced elsewhere, like in https://github.com/openedx/edx-platform/pull/33680. 
- I also noticed that we don't have things published in readthedocs, and created the issue: https://github.com/openedx/edx-drf-extensions/issues/403

Please mark resolved, unless you want to respond. Thank you.

docs/decisions/0002-remove-use-jwt-cookie-header.rst
```diff
@@ -21,14 +21,16 @@ Use of this header has several problems:
 Decision
 --------
 
-Replace the `HTTP_USE_JWT_COOKIE` header with forgiving authentication when using JWT cookies. By "forgiving", we mean that JWT authentication would no longer raise exceptions for failed authentication when using JWT cookies, but instead would simply return None.
+Replace the ``HTTP_USE_JWT_COOKIE`` header with forgiving authentication when using JWT cookies. By "forgiving", we mean that JWT authentication would no longer raise exceptions for failed authentication when using JWT cookies, but instead would simply return None.
 
 By returning None from JwtAuthentication, rather than raising an authentication failure, we enable services to move on to other classes, like SessionAuthentication, rather than aborting the authentication process. Failure messages could still be surfaced using `set_custom_metric` for debugging purposes.
```
feanil, line 26: Is it worth adding that this is how DRF expects its auth classes to behave?

feanil (approved):
It's a thing of beauty.  👨🏾‍🍳 

I had one small suggestion for the ADR but I don't consider it blocking.

docs/decisions/0002-remove-use-jwt-cookie-header.rst
robrap: Here are the DRF docs for custom authentication: https://www.django-rest-framework.org/api-guide/authentication/#custom-authentication

> To implement a custom authentication scheme, subclass `BaseAuthentication` and override the `.authenticate(self, request)` method. The method should return a two-tuple of `(user, auth)` if authentication succeeds, or `None` otherwise.
>
> In some circumstances instead of returning None, you may want to raise an `AuthenticationFailed` exception from the .authenticate() method.

The start of their custom authentication docs make it seem like you should return None by default, but maybe raise an exception under certain circumstances.

However, I was more familiar with the next statements, which practically state the reverse:

> Typically the approach you should take is:
> - If authentication is not attempted, return `None`. Any other authentication schemes also in use will still be checked.
> - If authentication is attempted but fails, raise an `AuthenticationFailed` exception. An error response will be returned immediately, regardless of any permissions checks, and without checking any other authentication schemes.

This latter section is what I took to be their actual recommendation. Since we actually do check and fail first, we are somewhat going against this DRF "typical approach" by returning None anyway and allowing other authentication methods to be tried.

Given this context, if you still have a recommended change, do you mind sending your proposed text to clarify? Thanks.

feanil: Gotcha, yea, that does make it confusing.  I think given the confusion in the DRF text, what you have now is good.

feanil, line 26:
```suggestion
By returning None from JwtAuthentication, rather than raising an authentication failure, we enable services to move on to other classes, like SessionAuthentication, rather than aborting the authentication process. Failure messages could still be surfaced using ``set_custom_metric`` for debugging purposes.
```

feanil (approved):

docs/decisions/0002-remove-use-jwt-cookie-header.rst
robrap: I fixed this, and also updated the method name to `set_custom_attribute`. Thanks.

feanil (approved):

robrap:
@feanil: [request] This is ready for review again. I squashed the earlier commits and rebased again. Reviewing by commit just doesn't work. I did find that reviewing without whitespace makes it pretty simple to see changes: https://github.com/openedx/edx-drf-extensions/pull/404/files?diff=unified&w=1.

feanil (approved):

Issue #371, opened 2023-08-14
[DEPR]: USE-JWT-COOKIE header


robrap:
### Proposal Date

2023-08-14

### Target Ticket Acceptance Date

2023-08-14

### Earliest Open edX Named Release Without This Functionality

Quince - 2023-10

### Rationale

The USE-JWT-COOKIE header was used by MFE's to inform backends when JWT cookies should be used. It has some complexities as detailed in the following ADR proposing its removal. See docs/decisions/0002-remove-use-jwt-cookie-header.rst in https://github.com/openedx/edx-drf-extensions/pull/197 (which hasn't merged as-of this initial ticket write up).

The complexity causes confusion, and we'd like to simplify that.

### Removal

This PR introduces the replacement and ability to disable the old behavior:
- https://github.com/openedx/edx-drf-extensions/pull/197

This ticket details some of the follow-up cleanup work for full removal:
- https://github.com/openedx/edx-drf-extensions/issues/365

### Replacement

The PR https://github.com/openedx/edx-drf-extensions/pull/197 also introduces the replacement, which is something called "forgiving JWT cookies", where we accept JWT cookies on all requests (not just those with a special header), but if it fails authentication, we allow the endpoint to try other forms of authentication before giving up.

robrap:
Note: This should move to "Removing" once accepted, since this work is in-progress.

robrap:
A main part of this is complete. I updated the task list in the PR description regarding the long tail.

robrap:
Update: I'm working on the next task, trying to get services upgraded. Unfortunately this uncovered an issue that arose in one service (2U-specific), and we've ticketed and will look into this.[ Private ticket link](https://2u-internal.atlassian.net/browse/COSMO-215).

jristau1984:
@robrap the internal ticket has been moved to Done.

robrap:
**Update**: This ticket is mostly unblocked and the task list in this ticket has been updated to reflect next steps.

robrap:
**UPDATE:** After learning that you cannot drop `use-jet-cookie` from CORS_ALLOW_HEADERS in any backend before all the frontends stop sending this header, the rest of this removal got more complicated. The PR description task list has been updated appropriately. I'm not sure when and if this work will proceed.

robrap:
Note to self: 
1. This ticket needs to be updated to match the newer DEPR issue template.
[...]

PR #408, opened 2023-11-20
fix!: bug in JWT vs session user id compare

robrap:
**Description:**

Fixes a bug for any service other than the identity service
(LMS/CMS), where the session's local service user id would never
match the JWT LMS user id when compared.

- The custom attribute jwt_auth_mismatch_session_lms_user_id was
 renamed to jwt_auth_mismatch_session_lms_user_id to make this more
 clear.
- The setting EDX_DRF_EXTENSIONS[VERIFY_LMS_USER_ID_PROPERTY_NAME]
 was added to enable choosing the user object property that contains
 the LMS user id, if one exists. If this is set to None (the
 default), the check will use the lms_user_id property if it is
 found, and otherwise will skip this additional protection. In case
 of an unforeseen issue, use 'skip-check' to skip the check, even
 when there is an lms_user_id property.
- The custom attribute jwt_auth_get_lms_user_id_status was added to
 provide observability into the new functionality.

**BREAKING CHANGE**: The breaking change only affects services with
ENABLE_FORGIVING_JWT_COOKIES enabled. It now requires the new setting
VERIFY_LMS_USER_ID_PROPERTY_NAME to be set appropriately in order to
provide the existing Session vs JWT user id check. Note that only
LMS/CMS will likely need to set this value.

Part of DEPR:
https://github.com/openedx/edx-drf-extensions/issues/371

**Before merge checklist:**
- [x] For 2U: Set VERIFY_LMS_USER_ID_PROPERTY_NAME to 'id' in LMS.

**Merge checklist:**
- [x] All reviewers approved
- [x] CI build is green
- [x] Version bump if needed
- [x] Changelog record added
- [x] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,9 +19,9 @@ Fixed
 * **BREAKING CHANGE**: Fixes a bug for any service other than the identity service (LMS/CMS), where the session's local service user id would never match the JWT LMS user id when compared.
 
   * The custom attribute jwt_auth_mismatch_session_lms_user_id was renamed to jwt_auth_mismatch_session_lms_user_id to make this more clear.
-  * The setting EDX_DRF_EXTENSIONS[VERIFY_LMS_USER_ID_PROPERTY_NAME] was added to enable choosing the user object property that contains the lms_user_id, if one exists. If this is unset, a service will simply skip this additional protection.
+  * The setting EDX_DRF_EXTENSIONS[VERIFY_LMS_USER_ID_PROPERTY_NAME] was added to enable choosing the user object property that contains the LMS user id, if one exists. If this is set to None (the default), the check will use the lms_user_id property if it is found, and otherwise will skip this additional protection. In case of an unforeseen issue, use 'skip-check' to skip the check, even when there is an lms_user_id property.
   * The custom attribute jwt_auth_get_lms_user_id_status was added to provide observability into the new functionality.
-  * The breaking change only affects services with ENABLE_FORGIVING_JWT_COOKIES enabled. It now requires the new setting VERIFY_LMS_USER_ID_PROPERTY_NAME to be set in order to provide the existing Session vs JWT user id check.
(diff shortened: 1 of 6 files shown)


edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -351,25 +352,74 @@ def _is_jwt_cookie_and_session_user_mismatch(self, request, jwt_user_id):
...
+        #        - id-not-found (the property exists, but returned None)
+
+        lms_user_id_property_name = get_setting(VERIFY_LMS_USER_ID_PROPERTY_NAME)
+
+        # This special value acts like an emergency disable toggle in the event that the user object has an lms_user_id,
+        # but this LMS id check starts causing unforeseen issues and needs to be disabled.
+        skip_check_property_name = 'skip-check'
+        if lms_user_id_property_name == skip_check_property_name:
+            set_custom_attribute('jwt_auth_get_lms_user_id_status', skip_check_property_name)
+            return None
```
robrap, line 398: **Author Note:** It was taking me equally long trying to justify we we might not want this, and since I want to quickly get to removing the forgiven JWT toggle, I would have had no kill switch for this if it turns out to cause an issue. So, I decided to add this additional bit of complexity.

@feanil: Please mark as resolved if you have no response, or start a conversation. Thanks.

feanil (approved):

CHANGELOG.rst
```diff
@@ -12,6 +12,17 @@ Change Log
 Unreleased
 ----------
 
+[9.0.0] - 2023-11-27
+
+Fixed
+~~~~~
+* **BREAKING CHANGE**: Fixes a bug for any service other than the identity service (LMS/CMS), where the session's local service user id would never match the JWT LMS user id when compared.
+
+  * The custom attribute jwt_auth_mismatch_session_lms_user_id was renamed to jwt_auth_mismatch_session_lms_user_id to make this more clear.
```
timmc-edx, line 21: I think the first copy here is supposed to be `jwt_auth_mismatch_session_user_id`

robrap: Thanks. I created https://github.com/openedx/edx-drf-extensions/pull/411.

PR #414, opened 2023-12-05
fix: bug with service user id vs lms user id

robrap:
**Description:**

- Fixed bug for successful JWTs where the JWT user id was still
using the service user id, rather than the LMS user id, so comparison
against the LMS user id would fail.
- As part of the bug fix, the custom attribute
``failed_jwt_cookie_user_id`` was renamed to
``jwt_cookie_lms_user_id``, and will be set for all JWT cookies.
Since this is only a breaking change for recently added monitoring,
this won't be versioned as a breaking change.

This is part of:
https://github.com/edx/edx-arch-experiments/issues/429

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bump if needed
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/authentication.py
+++ b/edx_rest_framework_extensions/auth/jwt/authentication.py
@@ -124,9 +124,13 @@ class JwtAuthentication(JSONWebTokenAuthentication):
             # CSRF passed validation with authenticated user
 
             # adds additional monitoring for mismatches; and raises errors in certain cases
-            self._monitor_or_enforce_successful_jwt_cookie_and_session_user_mismatch(
-                request, jwt_user_id=user_and_auth[0].id
-            )
+            is_mismatch = self._is_jwt_cookie_and_session_user_mismatch(request)
+            if is_mismatch and get_setting(ENABLE_SET_REQUEST_USER_FOR_JWT_COOKIE):
+                raise JwtSessionUserMismatchError(
+                    'Failing otherwise successful JWT authentication due to session user mismatch '
+                    'with set request user.'
+                )
+
             set_custom_attribute('jwt_auth_result', 'success-cookie')
             return user_and_auth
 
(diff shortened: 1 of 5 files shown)


edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -371,6 +332,34 @@ def _is_jwt_cookie_and_session_user_mismatch(self, request, jwt_user_id):
...
+        """
+        Returns the LMS user id from the JWT cookie, or None if not found
+
+        Notes:
+        - Must only be called in the case of a JWT cookie.
+        - Also provides monitoring details for mismatches.
+        """
+        try:
+            cookie_token = JSONWebTokenAuthentication.get_token_from_cookies(request.COOKIES)
+            invalid_decoded_jwt = unsafe_jwt_decode_handler(cookie_token)
```
robrap, line 345: **Author note**:
1. Reviewing by commit may be simpler, because this method was moved in a separate commit.
2. We will now be decoding all JWT cookies, which hopefully won't be a performance issue since we aren't even verifying the signature or anything.

Review can mark as resolved once read.

edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -371,6 +332,34 @@ def _is_jwt_cookie_and_session_user_mismatch(self, request, jwt_user_id):
 
         return True
 
+    def _get_and_monitor_jwt_cookie_lms_user_id(self, request):
+        """
+        Returns the LMS user id from the JWT cookie, or None if not found
+
+        Notes:
+        - Must only be called in the case of a JWT cookie.
```
timmc-edx, line 340: This sentence used to end with "failure", but the removal makes it unclear to me. Maybe this?

```suggestion
        - Must only be called when a JWT cookie is present.
```

timmc-edx (approved):
One doc nit, but otherwise looks good.

edx_rest_framework_extensions/auth/jwt/authentication.py
robrap: I deleted the line instead. Thanks.

PR #433, opened 2024-02-01
fix: change base class for JwtAuthenticationError

syedsajjadkazmii:
**Description:**

**Issue:** When JwtAuthentication class [raises](https://github.com/openedx/edx-drf-extensions/blob/master/edx_rest_framework_extensions/auth/jwt/authentication.py#L113-L119) `JwtUserEmailMismatchError` custom exception, it is not identified by the default exception handler of rest framework, therefore, returns `response = None` and hence we [get](https://www.django-rest-framework.org/api-guide/exceptions/#custom-exception-handling) `500 Server error`. 

**Fix:** This PR handles this issue by subclassing `JwtAuthenticationError` from the `AuthenticationFailed` exception which is a known exception and will be handled by the default exception handler.

**JIRA:**

[VAN-1694](https://openedx.atlassian.net/browse/VAN-1694)

**Additional Details**

* **Dependencies:**: List dependencies on other outstanding PRs, issues, etc.
* **Merge deadline:** List merge deadline (if any)
* **Testing instructions:** Provide non-trivial testing instructions
* **Author concerns:** List any concerns about this PR

**Reviewers:**
- [x] tag reviewer

**Merge checklist:**
- [ ] All reviewers approved
- [ ] CI build is green
- [ ] Version bump if needed
- [ ] Changelog record added
- [ ] Commits are squashed

**Post merge:**
- [ ] Create a tag
      finished.

### DIFF
--- a/edx_rest_framework_extensions/auth/jwt/authentication.py
+++ b/edx_rest_framework_extensions/auth/jwt/authentication.py
@@ -24,7 +24,7 @@ from edx_rest_framework_extensions.settings import get_setting
 logger = logging.getLogger(__name__)
 
 
-class JwtAuthenticationError(Exception):
+class JwtAuthenticationError(exceptions.AuthenticationFailed):
     """
     Custom base class for all exceptions
     """
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,11 @@ Change Log
 Unreleased
 ----------
 
+[10.2.0] - 2024-01-26
+---------------------
+
+* Subclasses base exception ``JwtAuthenticationError`` from Django Rest Framework's ``AuthenticationFailed`` exception, allowing ``JwtAuthenticationError`` to be recognized and handled by the default exception handler of the Django Rest Framework.
+
 [10.1.0] - 2024-01-26
 ---------------------
 
--- a/edx_rest_framework_extensions/__init__.py
+++ b/edx_rest_framework_extensions/__init__.py
@@ -1,3 +1,3 @@
 """ edx Django REST Framework extensions. """
 
-__version__ = '10.1.0'  # pragma: no cover
+__version__ = '10.2.0'  # pragma: no cover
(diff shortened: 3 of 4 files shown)


syedsajjadkazmii:
Hi @robrap, could i please get your review here?

edx_rest_framework_extensions/auth/jwt/authentication.py
```diff
@@ -34,7 +34,7 @@ class JwtSessionUserMismatchError(JwtAuthenticationError):
     pass
 
 
-class JwtUserEmailMismatchError(JwtAuthenticationError):
+class JwtUserEmailMismatchError(exceptions.AuthenticationFailed):
```
robrap, line 37: 1. Could we change the base class instead?
```
class JwtAuthenticationError(exceptions.AuthenticationFailed):
```
2. Is there any additional assertion we could add to a test to ensure that the right type of exception is being used?

syedsajjadkazmii: > Could we change the base class instead?

Yes, done.

> Is there any additional assertion we could add to a test to ensure that the right type of exception is being used?

Since `AuthenticationFailed` is a known exception to the default exception handler therefore it gets handled by the exception handler and we only get a `401 Unauthorized` response. Exception is not being propagated up and we are not getting it in the test. So, I don't think we can assert the exception type.

robrap: Although it is unlikely that someone will break this, a unit test that creates an instance of `JwtSessionUserMismatchError` and asserts that it is also an instance of `AuthenticationFailed` would prevent regressions. I'm not sure if it is worth it, but it is also very quick to implement. I'll leave it to you to decide.

robrap (approved):
I left a unit test idea, but I'll leave that to you to decide. See https://github.com/openedx/edx-drf-extensions/pull/433#discussion_r1476173044. Thanks.

syedsajjadkazmii:
> I left a unit test idea, but I'll leave that to you to decide. See [#433 (comment)](https://github.com/openedx/edx-drf-extensions/pull/433#discussion_r1476173044). Thanks.

Thanks, I have added the test

Issue #481, opened 2024-09-10
Add Support for Python 3.12


irtazaakram:
**Description:**

Follow and complete all of the following steps for the issue to be marked as done.

- [x] Update the `tox.ini`, `setup.py`, `setup.cfg`, or `pyproject.toml` to add Python 3.12 to the supported versions list.
- [x] Ensure GitHub Actions workflows, particularly the CI matrix, reflect Python 3.12 as a supported version.
- [x] Review and adjust any build or testing configurations to ensure compatibility with Python 3.12.
- [x] Run all tests and verify they pass in the CI with Python 3.12.
- [x] Add a CHANGELOG entry: “Added support for Python 3.12.”
- [x] Bump the package version and release a new version on GitHub & PyPI (if applicable).

iamsobanjaved:
PR to drop support for Python 3.8
- https://github.com/openedx/edx-drf-extensions/pull/474

@huniafatima-arbi kindly complete if any item is pending from the checklist otherwise close the issue.

PR #488, opened 2024-10-21
feat: add python 3.12 support

BilalQamar95:
## Description

For details, see tracking issue https://github.com/openedx/edx-drf-extensions/issues/481


CHANGELOG.rst
```diff
@@ -12,6 +12,10 @@ Change Log
 Unreleased
 ----------
 
+[10.5.0] - 2024-10-21
+---------------------
+* Added support for  ``Python 3.12``
```
UsamaSadiq, line 17: Let's change it to following instead.
```suggestion
* Upgraded to use ``Python 3.12``
```

edx_rest_framework_extensions/auth/jwt/decoder.py
```diff
@@ -267,6 +267,7 @@ def _verify_jwt_signature(token, jwt_issuer, decode_symmetric_token):
         raise jwt.InvalidTokenError(exc_info[2]) from token_error
 
 
+# pylint: disable=R0917
```
UsamaSadiq, line 270: Is this one still needed after updating pylintrc?

.github/workflows/ci.yml
```diff
@@ -13,8 +13,8 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.11']
-        toxenv: [quality, docs, django42-drflatest]
+        python-version: ['3.12']
+        toxenv: [quality, docs, django42-drflatest, django42]
```
UsamaSadiq, line 17: We already have `django42-drflatest` target so we don't need to add another `django42` target.

UsamaSadiq (changes requested):

edx_rest_framework_extensions/auth/jwt/decoder.py
BilalQamar95: Yes, it suppresses `too-many-positional-arguments` error. 

Screenshot 2024-10-23 at 8 30 38 AM

UsamaSadiq: Let's add `too-many-positional-arguments` in `pylintrc_tweaks` disable section instead of within the file.

UsamaSadiq (approved):
Let's squash the commits before merging the PR and remember to release new package version.

PR #528, opened 2025-07-14
fix: Expand DRF permission composition support to JwtRedirectToLoginIfUnauthenticatedMiddleware

pwnage101:
IDAs which enable this middleware should still be able to use DRF permission composition. This PR reinforces the stated promise that this middleware is a no-op for any ViewSet that does not actually use LoginRedirectIfUnauthenticated.


pwnage101:
hm, who can review and has write access?

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -35,33 +35,34 @@
 log = logging.getLogger(__name__)
 
 
+def _iter_included_base_classes(view_permissions):
```
robrap, line 38: Did this move unchanged? Took a bit to realize that this was not part of the fix.

edx_rest_framework_extensions/auth/jwt/middleware.py
```diff
@@ -161,8 +162,11 @@ def _check_and_cache_login_required_found(self, view_func):
         Checks for LoginRedirectIfUnauthenticated permission and caches the result.
         """
         view_class = _get_view_class(view_func)
-        view_permission_classes = getattr(view_class, 'permission_classes', tuple())
-        is_login_required_found = _includes_base_class(view_permission_classes, LoginRedirectIfUnauthenticated)
+        view_permissions = getattr(view_class, 'permission_classes', tuple())
```
robrap, line 165: Nit: Not really introduced by you, but above we have:
```
view_permissions = list(getattr(view_class, 'permission_classes', []))
```
Is there a reason for the inconsistency?

robrap (commented):
Can we get a unit test that requires your fix to pass?

edx_rest_framework_extensions/auth/jwt/middleware.py
pwnage101: Yep, I moved it out of the class to re-use in a different class, but did not modify the logic.

edx_rest_framework_extensions/auth/jwt/middleware.py
pwnage101: It looks like an older version of this code used to iterate over the permission_classes using `list.pop()`:

https://github.com/pwnage101/edx-drf-extensions/blob/1d470a29a16bcaf5c9247c8236def8a4be56b077/edx_rest_framework_extensions/auth/jwt/middleware.py#L56-L57

The recursive helper function _iter_included_base_classes appears to support any iterable, so it shouldn't matter whether a list or tuple is passed. That said, I'm happy to make it consistently a list.

robrap:
@pwnage101: Would it make sense to create an issue in this repo describing the bug and pointing to this PR as a potential fix? Would you mind taking care of that? It could even get a "good first issue" label, since you handled much of it already.

pwnage101:
done: https://github.com/openedx/edx-drf-extensions/issues/533

robrap:
Thank you.

